PR #48, opened 2015-01-09
Add a simple shared key encoder and make configuration more flexible

dbestevez:
Hi,

If you create a custom JWT encoder you still have to generate and configure SSH keys even if the encoder does not need them.

What this PR does:
- Allow to choose the signer (from [`namshi/JOSE`](https://github.com/namshi/jose) library) to use by the encoder.
- Make `encoder_service` parameter required and move key configuration to service definition.
- Add a simple shared-key JWT encoder to use with HMAC signer.

Thanks.


slashfan:
Hi,

Very nice work ! One question though, why not provide encoder services right out of the box and make them configurable through the configuration extension ? This would make the bundle more user friendly.

dbestevez:
You are right but the problem (actually my problem) is that the default JWTEncoder depends on the SSH key files. If you want to provide a default configuration through the configuration extension you force the users to generate those files even if they are going to use a custom encoder.

Maybe we can change the JWTEncoder to also allow strings to use as keys so we can provide them throught the configuration extension.

What do you think? Is there a better way to remove that constraint and keep the encoder configurable through the configuration extension?

bartv2:
Any news on this? My project doesn't need RS256 and i would like to use HS256 without having to use a custom dependency.

urosgruber:
+1

MarcelHernandez:
+1

PR #55, opened 2015-03-23
Use the recommended way to install Composer pkgs

willdurand:


slashfan:
You're right, this was written a long time ago ! Thanks !

willdurand:
thanks!

PR #67, opened 2015-05-13
Move security details to parameters.yml.dist

Maltronic:
As per http://symfony.com/doc/current/best_practices/configuration.html infrastructure parameters should go in the parameters.yml file rather than config.yml, this also enables different token lifetimes and key passphrases for development / production environments.


slashfan:
Nice catch !

PR #70, opened 2015-06-05
fixed deprecated errors for symfony 2.6 plus

slashfan:
Fix deprecated warnings and errors during tests for symfony 2.6+.
Used same logic as in FOSUserBundle to handle both SecurityContextInterface and TokenStorageInterface.


slashfan:
Tests are all green, should not introduce any BC break. Is it ok to merge ?

PR #71, opened 2015-06-05
Fixing a missing use statement for Reference

adetwiler:
* Use statement was preventing the bundle from properly working
PHP Fatal error:  Class 'Lexik\Bundle\JWTAuthenticationBundle\DependencyInjection\Reference' not found in /vagrant/vendor/lexik/jwt-authentication-bundle/Lexik/Bundle/JWTAuthenticationBundle/DependencyInjection/LexikJWTAuthenticationExtension.php on line 43


slashfan:
Nice catch thanks !

PR #72, opened 2015-06-18
Add a curl sample to get the token

Soullivaneuh:
I made some time to find how to get the token with a simple `curl` command before guessing it on the testing cookbook.

I propose to add it on the main documentation to get future user life easier. :+1:


slashfan:
Thanks !

PR #73, opened 2015-07-04
add JWTEncodedEvent so JWT string is available after its creation

9orky:
Sometimes this is useful (or even essential) to get JWT string right after it is created. This PR introduces new type of event - JWTEncodedEvent which is being dispatched right after JWT is created (encoded) and contains JWT string.

### DIFF
--- a/Event/JWTEncodedEvent.php
+++ b/Event/JWTEncodedEvent.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Lexik\Bundle\JWTAuthenticationBundle\Event;
+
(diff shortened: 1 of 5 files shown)


fesor:
Can you provide use cases?

9orky:
Of course :-) We need to persist JWT so it's possible to check if user's token is valid. To avoid hitting the backend this check will be done from varnish level. So if we want to store JWT string representation we need to know its value when it's created.

slashfan:
Nice addition. Ok for merge.

PR #74, opened 2015-07-06
Update composer.json

Irvyne:
Pass namshi/jose from v2 to v5


slashfan:
Hi, what are the benefits of the version 5 over 2 ? Are you sure it doesn't introduce any BC break ?

PR #76, opened 2015-07-21
Add cookie token extractor

tnucera:
Hi,

We store the token in a cookie (secure and httponly) in onAuthenticationSuccess like this :
<pre>
    public function onAuthenticationSuccess(Request $request, TokenInterface $token)
    {
        $response = parent::onAuthenticationSuccess($request, $token);

        $tokenJWT = json_decode($response->getContent(), true)['token'];

        // Ajoute le token avec l'id de l'utilisateur en clé dans les Redis configurés, avec le ttl contenu dans la conf
        $this->rms->set($this->redisDB, $token->getUser()->getId(), $tokenJWT, $this->jwtTokenTTL);

        // Crée le cookie contenant le token, avec le ttl contenu dans la conf
        $response->headers->setCookie(new Cookie('CERBERE', $tokenJWT, (new \DateTime())->add(new \DateInterval('PT' . $this->jwtTokenTTL . 'S')), '/', null, $this->cookieSecure));

        return $response;
    }
</pre>

Cookie is send transparently from Front to Back.
So we need a cookie token extractor.

Thanks

### DIFF
--- a/Tests/TokenExtractor/CookieTokenExtractorTest.php
+++ b/Tests/TokenExtractor/CookieTokenExtractorTest.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace Lexik\Bundle\JWTAuthenticationBundle\Tests\TokenExtractor;
+
+use Lexik\Bundle\JWTAuthenticationBundle\TokenExtractor\CookieTokenExtractor;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * CookieTokenExtractorTest
+ *
+ * @author Nicolas Cabot <n.cabot@lexik.fr>
+ */
+class CookieTokenExtractorTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * test getRequestToken
+     */
+    public function testGetTokenRequest()
+    {
+        $extractor = new CookieTokenExtractor('BEARER');
+
+        $request = new Request();
+        $this->assertFalse($extractor->extract($request));
+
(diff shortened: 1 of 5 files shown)


slashfan:
Hi,

Nice addition ! I'm waiting for the travis build to complete but it seems good for merging. 
2 little details though. Could you squash you commits into a single one ? And what do you think about giving the default cookie name the same as the query parameter ? 

Thanks.

Tests/TokenExtractor/CookieTokenExtractorTest.php
```diff
@@ -0,0 +1,33 @@
...
+     */
+    public function testGetTokenRequest()
+    {
+        $extractor = new CookieTokenExtractor('Bearer');
+
+        $request = new Request();
+        $this->assertFalse($extractor->extract($request));
+
+        $request = new Request();
+        $request->cookies->add(['Bear' => 'testtoken']);
```
slashfan: Please use classic array syntax for PHP 5.3 compatibility.

tnucera:
Ok it's done.
For the same name as the query parameter, it's more a thinking to have about the global conf of the bundle.
Indeed, there is code duplication in the factory at this level.
Why not just have (or something like this) :
<pre>
            extractors:
                enabled: ['authorization_header', 'cookie', 'query_parameter']
                name:     Bearer
</pre>

slashfan:
That's a good idea. In the mean time, let's merge this PR.

PR #88, opened 2015-08-28
Add WWW-Authenticate response header on 401

teohhanhui:


slashfan:
Hi, would you care to elaborate ?

teohhanhui:
http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.2

> 10.4.2 401 Unauthorized
>
> The request requires user authentication. The response MUST include a WWW-Authenticate header field (section 14.47) containing a challenge applicable to the requested resource.

teohhanhui:
Actually, as far as I can tell, the "Bearer" part is standard (it's the same even for OAuth 2). It's only what follows that might vary. But we don't necessarily have to provide that flexibility for now... (Perhaps another PR for further enhancement?)

slashfan:
I think you're right. Let's merge that for the moment.

Issue #89, opened 2015-08-31
Stoping propagation of `AUTHENTICATION_SUCCESS` before dispatching it.


pawaclawczyk:
Here https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Security/Http/Authentication/AuthenticationSuccessHandler.php#L52 is called `setResponse` method which internally calls `Symfony\Component\HttpKernel\Event\GetResponseEvent::stopPropagation` then the event is dispatched and stopped on first listener it reaches.

Is this expected beahavior?

slashfan:
Hello.

This is indeed not the way it should work and it is not the way it worked at the beginning of the project. 
I'll try to fix it, in the meantime feel free to make a PR.

PR #92, opened 2015-09-21
Fix authentication event propagation

mRoca:
This PR fixes https://github.com/lexik/LexikJWTAuthenticationBundle/issues/89 .

I've removed the `extends GetResponseEvent` from events : this parent event is bound to the kernel (see the [symfony api doc](http://api.symfony.com/2.7/Symfony/Component/HttpKernel/Event/GetResponseEvent.html)), and does not make sense here. Moreover, the [`GetResponseEvent::setResponse()`](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/a0e2f0a0fafdd41ccdba1c368101c242f2463918/Security/Http/Authentication/AuthenticationSuccessHandler.php#L52) method does a `stopPropagation()` and prevents the use of more than one listener.

### DIFF
--- a/Event/AuthenticationFailureEvent.php
+++ b/Event/AuthenticationFailureEvent.php
@@ -2,8 +2,9 @@
 
 namespace Lexik\Bundle\JWTAuthenticationBundle\Event;
 
+use Symfony\Component\EventDispatcher\Event;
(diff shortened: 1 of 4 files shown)


markitosgv:
+1 to this fix.

PR #96, opened 2015-10-06
Add custom authorization header name

pdoreau:
Hi, I need double authentication : HTTP basic for general restriction and JWT at API level.
They both used the same header (Authorization).
This PR allows to use a custom header name for JWT token in order to handle double authentication.

### DIFF
--- a/Tests/TokenExtractor/AuthorizationHeaderTokenExtractorTest.php
+++ b/Tests/TokenExtractor/AuthorizationHeaderTokenExtractorTest.php
@@ -17,13 +17,17 @@ class AuthorizationHeaderTokenExtractorTest extends \PHPUnit_Framework_TestCase
      */
     public function testGetTokenRequest()
     {
-        $extractor = new AuthorizationHeaderTokenExtractor('Bearer');
+        $extractor = new AuthorizationHeaderTokenExtractor('Bearer', 'Authorization');
 
         $request = new Request();
         $this->assertFalse($extractor->extract($request));
 
         $request = new Request();
         $request->headers->set('Authorization', 'Bear testtoken');
+        $this->assertFalse($extractor->extract($request));
+
+        $request = new Request();
+        $request->headers->set('Authorizat', 'Bearer testtoken');
         $this->assertFalse($extractor->extract($request));
 
         $request = new Request();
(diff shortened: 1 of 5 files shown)


TokenExtractor/AuthorizationHeaderTokenExtractor.php
```diff
@@ -17,11 +17,18 @@ class AuthorizationHeaderTokenExtractor implements TokenExtractorInterface
...
+     * @var string
+     */
+    protected $name;
+
+    /**
+     * @param string $pr
+     * @param string $name
      */
-    public function __construct($prefix)
+    public function __construct($prefix, $name)
```
slashfan: It would be safer to set a default value of 'Autorization' for the $name argument, to prevent any BC break.

slashfan:
Hi, nice addition for special cases.

TokenExtractor/AuthorizationHeaderTokenExtractor.php
pdoreau: Default $name is set in JWTFactory.php, the same way default $prefix is defined

Tests/TokenExtractor/AuthorizationHeaderTokenExtractorTest.php
```diff
@@ -27,6 +27,10 @@ public function testGetTokenRequest()
         $this->assertFalse($extractor->extract($request));
 
         $request = new Request();
+        $request->headers->set('Authorizatio', 'Bearer testtoken');
```
markitosgv: Is this ok? Or it must be:

```php
$request->headers->set('Authorization', 'Bearer testtoken');
```

TokenExtractor/AuthorizationHeaderTokenExtractor.php
```diff
@@ -17,11 +17,18 @@ class AuthorizationHeaderTokenExtractor implements TokenExtractorInterface
...
+     * @var string
+     */
+    protected $name;
+
+    /**
      * @param string $prefix
+     * @param string $name
      */
-    public function __construct($prefix)
+    public function __construct($prefix, $name)
```
slashfan: Again, I think there should be a default value for the $name parameter, even if it is overwritable by the dependency injector like any other argument. The class is called AuthorizationHeaderTokenExtractor, so it would make sense to have it "preconfigured" for this header name. 

Maybe a constant DEFAULT_HEADER_NAME = 'Authorization' and a constructor like __construct($prefix, $name = self::DEFAULT_HEADER_NAME).

In a future major version, the class could be renamed to HeaderTokenExtractor.

PR #109, opened 2015-11-09
Updated namshi version

TiS:
Updated namsho version to allow newer phpseclib.

All of the phpunit tests passed, so I guess it's fine.


TiS:
Closed - I have to make more updates first. Another will come.

PR #110, opened 2015-11-09
Updated to newest namshi/jose. Dropped support for PHP 5.3

TiS:
Updated namsho version to allow newer phpseclib.

All of the phpunit tests passed, so I guess it's fine.

This drops the support of PHP 5.3, but I think it's worth it, as more and more libraries use phpseclib 2.0. This resolves conflicts eg with newest google API client library.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-    - 5.3
     - 5.4
     - 5.5
     - 5.6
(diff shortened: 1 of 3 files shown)


tristanbes:
:+1:

Spomky:
PHP 5.4 is not supported anymore, you can drop it too.

TiS:
Unsupported, but still used on many servers, so I will be carefull with this one (it's best to stick with lowest feasible version).

ceednee:
:+1:

sdieunidou:
:+1:

slashfan:
Hi @TiS sorry for the delay, could you update the PR ?

PR #117, opened 2015-12-14
Allow empty ttl

soyuka:
Actually `namshi/jose` does not prevent an empty ttl. Can be useful in some cases instead of setting a really long one.


slashfan:
Hi, 

Indeed the ttl claim is optional https://self-issued.info/docs/draft-ietf-oauth-json-web-token.html#expDef

Thanks !

PR #126, opened 2016-01-17
Use requestStack instead of request

SmurfyFR:
Hello,

With Symfony 3.0(.1), the Request service has been once for all removed from the service container. This service has been replaced by the request_stack service.

The problem was that the JWTManager is relying on the request service and as this service has been removed...

I don't think it's possible to keep a backward-comptability with SF < 2.4 (ie: SF 2.3, which is a LTS ending in May 2016 ...), so maybe creating a tag could be a workaround?

I remain at your disposal for any further modifications

Regards,

PS : Sorry for the double PR, I made mistakes trying to correct a comment type ...

### DIFF
--- a/Services/JWTManager.php
+++ b/Services/JWTManager.php
@@ -134,14 +134,14 @@ class JWTManager implements JWTManagerInterface
     /**
      * @param RequestStack|Request $requestStack
      */
-    public function setRequest($requestStack = null)
+    public function setRequest($requestStack)
     {
-        if (!$requestStack instanceof Request && !$requestStack instanceof RequestStack) {
-            throw new \InvalidArgumentException('Argument 1 should be an instance of Symfony\Component\HttpFoundation\RequestStack or Symfony\Component\HttpFoundation\Request');
+        if ($requestStack instanceof Request) {
+            $this->request = $requestStack;
+        } elseif ($requestStack instanceof RequestStack) {
+            $this->request = $requestStack->getCurrentRequest();
         } else {
-            if($requestStack instanceof RequestStack) {
-                $this->request = $requestStack instanceof RequestStack ? $requestStack->getCurrentRequest() : $requestStack;
-            }
+            throw new \InvalidArgumentException('Argument 1 should be an instance of Symfony\Component\HttpFoundation\RequestStack or Symfony\Component\HttpFoundation\Request');
         }
     }
 
--- a/Services/JWTManager.php
+++ b/Services/JWTManager.php
@@ -8,7 +8,7 @@ use Lexik\Bundle\JWTAuthenticationBundle\Event\JWTDecodedEvent;
 use Lexik\Bundle\JWTAuthenticationBundle\Event\JWTEncodedEvent;
 use Lexik\Bundle\JWTAuthenticationBundle\Events;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
-use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
 use Symfony\Component\Security\Core\User\UserInterface;
 use Symfony\Component\PropertyAccess\PropertyAccess;
@@ -131,11 +131,11 @@ class JWTManager implements JWTManagerInterface
     }
 
     /**
-     * @param Request $request
+     * @param RequestStack $requestStack
      */
-    public function setRequest(Request $request = null)
+    public function setRequest(RequestStack $requestStack = null)
     {
-        $this->request = $request;
+        $this->request = $requestStack ? $requestStack->getCurrentRequest() : null;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


JHGitty:
:+1: +1

slashfan:
Hi, thanks for the PR ! 

Would you or anyone be willing to work on a "hack" to try to make it both symfony 2.3+ and 3.0+ compatible before merging ? Like we did with the security storage / security context deprecation ?

I'd like to keep the 1.* branch alive until symfony 2.3 end of life in may. The 2.0 branch will arrive after that, symfony 2.7+ and 3.0+ compatible, allowing to make a big refactoring and cleanup of the code as well as new features.

SmurfyFR:
Just pushed the "hack" ;)

Services/JWTManager.php
```diff
@@ -131,11 +132,17 @@ public function setUserIdentityField($userIdentityField)
...
      */
-    public function setRequest(Request $request = null)
+    public function setRequest($requestStack = null)
     {
-        $this->request = $request;
+        if (!$requestStack instanceof Request && !$requestStack instanceof RequestStack) {
+            throw new \InvalidArgumentException('Argument 1 should be an instance of Symfony\Component\HttpFoundation\RequestStack or Symfony\Component\HttpFoundation\Request');
+        } else {
+            if($requestStack instanceof RequestStack) {
+                $this->request = $requestStack instanceof RequestStack ? $requestStack->getCurrentRequest() : $requestStack;
```
Spomky:
```php
if ($requestStack instanceof Request) {
    $this->request = $requestStack;
} elseif ($requestStack instanceof RequestStack) {
    $this->request = $requestStack->getCurrentRequest();
} else {
    throw new \InvalidArgumentException('Argument 1 should be an instance of Symfony\Component\HttpFoundation\RequestStack or Symfony\Component\HttpFoundation\Request');
}
```

Services/JWTManager.php
```diff
@@ -131,11 +132,17 @@ public function setUserIdentityField($userIdentityField)
     }
 
     /**
-     * @param Request $request
+     * @param RequestStack|Request $requestStack
      */
-    public function setRequest(Request $request = null)
+    public function setRequest($requestStack = null)
```
Spomky: `= null` is not need as you expect an object.

Spomky:
A better way could be to use a compiler pass and set the request/request_stack in the constructor.

SmurfyFR:
Done

But I don't know how to "to use a compiler pass". I just did what @slashfan said (do something like it has been done with the token_storage service.

Regards,

Spomky:
IMHO, the token_storage should also be injected with a compiler pass, but it is another topic.

1: Create the compiler pass in folder `DependencyInjection\Compiler\RequestCompilerPass.php`:

```php
<?php

namespace Lexik\Bundle\JWTAuthenticationBundle\DependencyInjection\Compiler;

use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
use Symfony\Component\DependencyInjection\ContainerBuilder;
use Symfony\Component\DependencyInjection\Reference;

final class RequestCompilerPass implements CompilerPassInterface
{
    /**
     * {@inheritdoc}
     */
    public function process(ContainerBuilder $container)
    {
        if (!$container->hasDefinition('lexik_jwt_authentication.jwt_manager')) {
            return;
        }

        $definition = $container->getDefinition('lexik_jwt_authentication.jwt_manager');

        if ($container->hasDefinition('request_stack')) {
            $definition->addMethodCall('setRequest', array(new Reference('request_stack')));
        } else {
            $definition->addMethodCall('setRequest', array(new Reference('request')));
        }
    }
}
```

2: Enable it in the bundle `Lexik\Bundle\JWTAuthenticationBundle\LexikJWTAuthenticationBundle`:

```php
...
use Lexik\Bundle\JWTAuthenticationBundle\DependencyInjection\Compiler\RequestCompilerPass;
...
class LexikJWTAuthenticationBundle extends Bundle
{
    ...
    public function build(ContainerBuilder $container)
    {
        parent::build($container);
        ...
        $container->addCompilerPass(new RequestCompilerPass());
    }
...
```

3: Remove the method call in the service definition

```xml
<call method="setRequest">
    <argument /> <!-- request_stack or request for Symfony <2.4 -->
</call>
```

Spomky:
I think the `composer.json` file should be updated too as tests did not passed because of missing components:

```json
"require": {
    "php": ">=5.4.8",
    "symfony/framework-bundle": "~2.3|~3.0",
    "symfony/security-bundle": "~2.3|~3.0",
    "symfony/console": "~2.3|~3.0",
    "namshi/jose": "~6.0"
},
```

Plus PHPUnit 5 could be used:
```php
"require-dev": {
    "phpunit/phpunit": "^4.1|^5.0",
    "symfony/phpunit-bridge": "~2.7|~3.0"
}
```

SmurfyFR:
Thanks. It's done :+1:

slashfan:
Great work guys !

Issue #128, opened 2016-01-22
JWTCreatedEvent Request is null


blt909:
Hi,

Since we've upgraded to 1.4.2, our listener onJWTCreated isn't working anymore.
It seems that $event->getRequest() is always null.
Our app is running on Symfony 2.8.2.

Here's our service definition : 
```yaml
# JWT Creation Listener
    pm.event.jwt_created_listener:
        class: AppBundle\EventListener\JWTCreatedListener
        arguments:
            - "@security.token_storage"
        tags:
            - { name: kernel.event_listener, event: lexik_jwt_authentication.on_jwt_created, method: onJWTCreated }
```

And the method handling the event : 
```php
public function onJWTCreated(JWTCreatedEvent $event)
    {
        $request = $event->getRequest();
        if (!$request) {
            return;
        }

        $user = $this->token_storage->getToken()->getUser();

        $payload       = $event->getData();
        $payload['pseudo']= $user->getPseudo();        

        $event->setData($payload);
    }
```

Am I missing somethig?

Thanks

slashfan:
Hi, 

This is indeed not the expected behavior. A big change on the request injection has been introduced with the 1.4.* tags to maintain both symfony 2.3+ and 3.0 compatibility, we'll have to look into it. But you're right the request should be available at runtime (although the $event->getRequest() as always been nullable). 

In the meantime, your handler doesn't seem to use the request anyway, so it should be safe to disable the check if you don't need it.

Spomky:
This is because the process is called outside the main request context.
As [stated here](http://symfony.com/blog/new-in-symfony-2-4-the-request-stack) `Outside the handling of a request, $requestStack->getCurrentRequest() returns null.`.

So to be sure to have the request always set, you must call `getMasterRequest()` instead of `getCurrentRequest()`.

slashfan:
Good point @Spomky anyone want to make a PR to fix this ?

Spomky:
I have just realized that there is not functional test of this feature. The expected PR should also add some tests by the way.

slashfan:
Exactly.

slashfan:
Ping ! Did anybody started to work on this one yet ?

slashfan:
Hi @blt909 do you still have the problem ?

PR #133, opened 2016-01-29
Always call for master request from request stack

stloyd:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixes     | #128
| License       | MIT

We had a lot of fatal errors that request is `null` after update to latest version of this bundle, this fixes all (at least from what we notice) of them.


slashfan:
Thanks !

PR #141, opened 2016-02-19
Update Namshi/jose to 6.1

jblanchon:
Hi,
Test seems to pass with this update,
Namshi/Jose 6.1 is using phpseclib 2.0.
Which is nice for using https://github.com/google/google-api-php-client combined with you bundle.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "symfony/framework-bundle": "~2.3|~3.0",
         "symfony/security-bundle": "~2.3|~3.0",
         "symfony/console": "~2.3|~3.0",
-        "namshi/jose": "~6.0"
+        "namshi/jose": "~6.1"
     },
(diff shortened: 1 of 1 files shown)


tristanbes:
Well, isn't it a problem if you bump PHP requirements to at least 5.5 to be compliant with Jose 6.1 ?

Right now we can't use Google api php client because of your bundle that require a previous version of Jose which requires phpspec lib <2.0. Yet google api php client requires phpsepct lib >=2

slashfan:
Hi, for now can't you just run `composer update lexik/jwt-authentication-bundle --with-dependencies` ? It should update the `namshi/jose` version to the latest 6.* tag (6.1.1 right now).

jblanchon:
Hi, I used your command, and it works fine, thank you for your help @slashfan

Issue #124, opened 2016-01-11
empty Response in case of invalid Token


digitalkaoz:
if im sending an invalid token i get an empty response, but i would expect some "Invalid Credentials" Response like i get when there is no token provided.

any idea what might be wrong i dont see an obvious error in my app...authentication works

SmurfyFR:
I'm getting the same problem here too.

slashfan:
Hi, 

It is indeed not the expected behavior. Have you looked at the [sandbox](https://github.com/slashfan/LexikJWTAuthenticationBundleSandbox) ? Specifically the [CLI testing section](https://github.com/slashfan/LexikJWTAuthenticationBundleSandbox#manually-check-token-generation-using-the-cli) ? If i submit a wrong password I receive a 401 with a code and message in the response body.

Could you show us your `security.yml ` ? Along with Symfony version and JWT bundle version you are using ?

digitalkaoz:
Hey using Symfony 3.0.1, 1.4.2 of your bundle, security looks as follows:

```yml
security:
    providers:
        in_memory:
            memory:
              users:
                  %api.user%:
                      password: %api.password%
                      roles: 'ROLE_USER'
    encoders:
        Symfony\Component\Security\Core\User\User: plaintext

    firewalls:
        dev:
            pattern: ^/(_(profiler|wdt)|css|images|js)/
            security: false

        doc:
            pattern: ^/$
            security: false

        token:
            pattern:  ^/token$
            stateless: true
            gfreeau_get_jwt: ~

        api:
            pattern:   ^/
            stateless: true
            lexik_jwt:
                query_parameter:
                    enabled: true
                    name:    token

    access_control:
        - { path: ^/login, roles: IS_AUTHENTICATED_ANONYMOUSLY }
        - { path: ^/_profiler, roles: IS_AUTHENTICATED_ANONYMOUSLY }
        - { path: ^/,       roles: IS_AUTHENTICATED_FULLY }

```

slashfan:
Hi @digitalkaoz did you find the cause of the problem ?

digitalkaoz:
Nope

Nicolas Cabot <notifications@github.com> schrieb am Mi., 27. Jan. 2016
18:57:

chalasr:
@slashfan To reproduce the problem, you need to send an invalid token, not submit a bad password. Try to make an authenticated request, and for example remove one or more characters in your token. You'll get an empty Response. Fixed in #142

[...]

PR #142, opened 2016-03-09
Add response message in case of invalid token

chalasr:
In response of issue #124 , in the JWTListener I replaced the Response by a JsonResponse (such as done by the [AuthenticationFailureHandler](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Security/Http/Authentication/AuthenticationFailureHandler.php#L39))  but using the exception's message.

The result is that when the token is found but invalid, the following response is returned :

```json
{
  "code": 401,
  "message": "Invalid JWT Token"
}
```

Instead of an empty response as it's done currently.

If this makes sense, I can adapt the JWTListenerTest to check the behavior in this specific case (invalid token).

### DIFF
--- a/Security/Firewall/JWTListener.php
+++ b/Security/Firewall/JWTListener.php
@@ -5,7 +5,7 @@ namespace Lexik\Bundle\JWTAuthenticationBundle\Security\Firewall;
 use Lexik\Bundle\JWTAuthenticationBundle\TokenExtractor\TokenExtractorInterface;
 use Lexik\Bundle\JWTAuthenticationBundle\Security\Authentication\Token\JWTUserToken;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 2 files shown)


slashfan:
Hi @chalasr it's a good idea if it fixes the the bad token and empty response issue. Could you update the tests or add new ones for this features ?

chalasr:
Hi @slashfan. No problem for the tests, I think verifiying only that the response is not empty. I'll add a check in the JWTListenerTest this week, please tell me if you have more insight or if it's not sufficient. I be back when it's done.

chalasr:
Hello @slashfan 

I updated the tests (added a check to see if the response message corresponds to the exception message).
Have a good week end.

slashfan:
Thanks @chalasr

PR #147, opened 2016-03-23
Made the public and private key paths not required…

ovidiumght:
I made the public and private key paths not required because they might be not used when defining custom encoders. We check for them at run time if using the default encoder. If you use the JWT encoder you just need a pass phrase but you still need to define those two config options with not used text.


slashfan:
Good idea, thanks !

PR #154, opened 2016-04-01
Add OpenSSLKeyLoader

chalasr:
Add a service that loads keys from parameters (key paths and passphrase as arguments of the service).
The service has been injected in the JWTEncoder rather than the 3 parameters, same for the method `checkOpenSSLConfig` that has been moved in the service and now takes no argument.

This also Improve checks for keys and their corresponding exception messages (One for file existing/reading problem, one for openssl loading, this for both public and private keys).

This avoid duplicated code (Command & Encoder) and make the bundle configuration easier to debug (centralised key loading logic).

### DIFF
--- a/Tests/Services/OpenSSLKeyLoaderTest.php
+++ b/Tests/Services/OpenSSLKeyLoaderTest.php
@@ -1,0 +1,89 @@
+<?php
+
+namespace Lexik\Bundle\JWTAuthenticationBundle\Tests\Services;
+
+use Lexik\Bundle\JWTAuthenticationBundle\Services\OpenSSLKeyLoader;
+
+/**
+ * OpenSSLKeyLoaderTest
+ *
+ * @author Robin Chalas <robin.chalas@gmail.com>
+ */
+class OpenSSLKeyLoaderTest extends \PHPUnit_Framework_TestCase
+{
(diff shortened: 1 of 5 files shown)


slashfan:
Good idea @chalasr ! Thanks !

slashfan:
Do you think you could add some tests for the OpenSSLKeyLoader class ?

chalasr:
@slashfan Yes of course, I jusy realize that I forgotten them. I tell you when it's ok.

chalasr:
@slashfan Unit tests added !

slashfan:
Great !

Tests/Services/OpenSSLKeyLoaderTest.php
```diff
@@ -0,0 +1,89 @@
...
+     * @expectedException        RuntimeException
+     * @expectedExceptionMessage Private key "private.pem" does not exist or is not readable.
+     */
+    public function testLoadUnreadablePrivateKey()
+    {
+        $this->keyLoader->loadKey('private');
+    }
+
+    /**
+     * Test load unreadable private key.
```
chalasr: Should be `* Test load invalid private key.`, same for the next but public.
It would be nice if you can fix it in a next commit @slashfan. 
Thanks by advance!

Issue #153, opened 2016-03-30
fosrest exception wrapper handler


stickyhands:
Im using the jwt auth bundle side by side with the fos rest bundle,
I created my own exception wrapper handler to format my errors the way I want.

But I cant figure out how to intercept `{"code":401,"message":"Invalid JWT Token"}` which is supposed to be a `Symfony\Component\Security\Core\Exception\AuthenticationException`

chalasr:
This seems to be not supported for now.

If this feature come to be planed, I think the logic way would be to listen on the  `on_authentication_failure` (like implemented for the `on_authentication_successs` event, see [EVENTS::AUTHENTICATION_SUCESS Customization])(https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Resources/doc/2-data-customization.md#eventsauthentication_success---add-public-data-to-the-jwt-response).

@slashfan What do you think about this ? Do you know more ?

slashfan:
Hi, I don't really have time right now but I'm it is indeed an important feature for an API. 
What's your opinion @chalasr ? How would you implement that ?

chalasr:
@slashfan After a lot of tries, I just come to make it works using events.
I have a ready patch, but I would like to discuss with you before submit a PR.

First, I created an `Events::AUTHENTICATION_FAILURE` with a `$message` property that represents the `message` field of the Response body (i.e. `Bad credentials` by default).
Users can create a listener and subscribe this event such as done for the `Events::AUTHENTICATION_SUCCESS`, then call `AuthenticationFailureEvent::setMessage('custom failure message')`. The primary goal is to allow users to subscribe to this event in order to do custom stuffs in their application, customizing the response is a kind of bonus, and not very risky because changing the response itself is not possible (no `AuthenticationFailureEvent::setResponse` method).

Then, to answer this issue and make user able to subscribe on the "Invalid JWT Token" (which is not an authentication from credentials but from token), my first idea was to simply pass the authenticationFailureHandler to our JWTListener and call onAuthenticationFailure (so dispatch the same `AUTHENTICATION_FAILURE` than on invalid username/password credentials).
It worked well, but the problem is that users can't make a difference between the authentication failure from credentials and the one from invalid token, and so cannot return a different response depending on the use case.

So I created a second event called `Events::JWT_INVALID`.
[...]

PR #157, opened 2016-04-05
Allow to set a custom response in case of authentication failure or invalid/not found token

chalasr:
| Q             | A    |
|---------------|------|
| Bug fix?      | yes  |
| New feature?  | yes  |
| BC breaks?    | no   |
| Fixed tickets | #153 |
| Tests pass?   | yes  |

This make developers able to subscribe on the  `AUTHENTICATION_FAILURE` / `INVALID_JWT` events.

The first is dispatched when the provider cannot retrieve/authenticate credentials sent to the login route.
The second (new) is dispatched when the listener cannot retrieve/authenticate a token on a secured route.

Tests and documentation have been updated.


slashfan:
OK to merge. Thanks !

tjaari:
👍

PR #159, opened 2016-04-08
Fix anonymous access by removing the AuthenticationCredentialsNotFoundException 

chalasr:
| Q             | A    |
|---------------|------|
| Bug fix?      | yes  |
| New feature?  | no |
| BC breaks?    | no   |
| Fixed tickets | no |
| Tests pass?   | yes  |

For now, we can't handle the case of no token found in request, so I revert a part of my previous #157.
The problem is that we can't know if the current request can be done anonymously.

@slashfan If you see any way to know if the current request can be authenticated anonymously or not (retrieve firewall config?), I could be able to add a check and throw the exception only if the request cannot be authenticated anonymously.

### DIFF
--- a/Resources/doc/2-data-customization.md
+++ b/Resources/doc/2-data-customization.md
@@ -232,7 +232,7 @@ public function onAuthenticationFailureResponse(AuthenticationFailureEvent $even
 
 #### Events::JWT_INVALID - customize the invalid token response
 
-By default, if the token is invalid or not set, the response is just a json containing the corresponding error message and a 401 status code, but you can set a custom response.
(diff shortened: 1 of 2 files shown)


slashfan:
This may be overkill but why not dispatch a "jwt not found" event and let the flow of the request continue ? It will only be dispatched on a jwt protected firewall so developers can listen to it only if they are interested.

chalasr:
@slashfan Separate invalid token and token not found in two events was my first thought, but it seemed overkill while I was able to dispatch the JWT_INVALID event for the two. 
Now I think you are right, that would be adapted to add a JWT_NOT_FOUND and dispatch it while letting the request continue (so keep the normal behavior and take the control of the Response).

I be back with the changes quickly!

Issue #132, opened 2016-01-29
Version 2.0 - Symfony 2.8 and Symfony 3.0+ compatible only


slashfan:
Hi,

With Symfony 2.3 support coming to an end in May and Symfony 2.8 introducing great new components such as Guard, it would be interesting to create a new major version and drop support for Symfony 2.7 and below. 

That way we could remove all the hack code that tried to maintain compatibility with old Symfony versions, and using using Guard authenticators we could simplify the security layer and provide an out of the box stateless replacement for the form_login.

Any thoughts on this ? It would be great to have your opinion.

Thanks !

-----------

Currently we have those tasks :

- [x] Remove deprecated calls and hacks which maintained compatibility with old symfony versions (#173)
- [x] Remove class parameters from Resource/config/services.xml (#178)
- [x] Provide a more flexible default encoder through supporting SecLib in addition of OpenSSL (#162)
- [x] Avoid injecting Request objects (#200, #201)
- [x] Add functional tests to test the different workflows : anonymous requests, events dispatched, response codes... (#184)
- [x] Use `AppBundle` as namespace for code examples in documentation (#243)
- [ ] Replace legacy security system by Guard Authenticators:
    - [x] JWT token authenticator (#184)
    - [ ] form-login authenticator (#237, closed for now)

Any idea is welcome.

lucascherifi:
:+1:

Spomky:
Hi,

I agree with you. This also implies for PHP 5.4 to be dropped (this version is not supported anymore).

I take the opportunity to introduce you my [jose library](https://github.com/Spomky-Labs/jose) and [bundle](https://github.com/Spomky-Labs/JoseBundle).
At this time these projects are under active developpement, but a stable release is expected by the end of this semester.
Despite the fact that the library is not stable, you will see that it already supports a lot of features, including JWE (encrypted JWT), JWK (keys)/JWKSet (key sets) from different sources (key files, certificates, urls...). All algorithms listed in the [RFC7518](http://tools.ietf.org/html/rfc7518) are supported.
I will be glad if you can take time to check these projects. [We can discuss on these projects](https://gitter.im/Spomky-Labs/jose) if you need more information.

weaverryan:
:+1: @slashfan I'd like to help with the goals of the next release, if you'll have me :). I agree completely with what your initial ideas. What's the next step?

slashfan:
[...]

PR #162, opened 2016-04-14
[Encoder] Handle OpenSSL/phpseclib engines and algorithms

chalasr:
| Q             | A    |
|---------------|------|
| Bug fix?      | no  |
| New feature?  | yes |
| BC breaks?    | yes  |
| Deprecations | yes |
| Fixed tickets | n/a |
| Tests pass?   | yes  |

After discussed in #132, I open this PR representing one of the first new features of the `v2.0`.

@slashfan I need your opinion about:
- Structure and naming,
I used `\Encoder\NamshiJOSE` to avoid possible conflicts between two bridges named `JOSE`, maybe you have a better name?
Same for the class names, maybe prefixing encoders by `JWT` would make more sense?

- Tests,
If you have any idea or suggestion about how write tests for the new encoders and BTW improve the existing (encoding), the current tests should not pass with the large feature added, I'll try to improve them (for the ones concerning encoders).

- Documentation,
I think that all can be done in the [configuration reference](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Resources/doc/1-configuration-reference.md). Do you see any other part that need to be updated?
Do we need to give a way to generate keys for phpseclib in addition of the openssl keys generation process? (I have a local work in progress containing a `GenerateKeysCommand` and the`OpenSSLKeyGenerator`+`SecLibKeyGenerator` services, I'd have your opinion about providing this feature)
I think to add something about the key loaders, but not sure about if it shouldn't stay internal.

Waiting for your suggestions.
Also I'll be glad to have the opinion of any developer interested by this changes.

Steps:

- [x] Ability to choose the encoder's encryption algorithm used by the encoder
- [x] Ability to choose the encoder's encryption engine between phpseclib and openssl (default)
- [x] Create a JWSProvider that manages Namshi\JOSE\JWS instances
- [x] Create a new DefaultEncoder that takes the JWSProvider service as argument
- [x] Create two base JWT[Decode/Encode]FailureException classes and their children
- [x] Throw the good exception depending on the fail's reason during `DefaultEncoder::[encode/decode]()`
- [x] Handle exceptions properly from the JWTProvider
- [x] Update the existing tests / add new ones (unit & functional)
- [x] Update the documentation
- [x] Rebase after merged #164  & update changelog
- [x] Add an UPGRADE file and document this feature

The v2 is coming!


slashfan:
Hi,

- structure and naming : 
+1 having the vendor name is mandatory to avoid possible conflicts.

- tests : 
I think we should only test the different encoders services initialization as the underlying encoding libraries are already tested.

- documentation : 
+1 for the configuration reference.
I'm not sure it's the bundle's work to provide keys generation logic, a simple explanation in the documentation should suffice. What do you think ?

Again, thanks for the hard work @chalasr !

chalasr:
@slashfan You're welcome, I'll personally benefit from this changes and it's 

For the keys generation, you're right, I suspected that it is the reason of why you doesn't provide it before. The reason of I worked on is that I needed it for test that SecLib encoder works properly.
I'll surely open it as a lite standalone library and maybe we could give a link for people that are not familiar with this kind of stuffs.

chalasr:
@slashfan You're welcome, I'll personally benefit from this changes, plus it's a pleasure to work on cool things.

For the keys generation, you're right, I suspected that it is the reason of why you doesn't provide it before. The reason of I worked on is that I needed it for test that SecLib encoder works properly.
I'll surely open it as a lite standalone library and maybe we could give a link for people that are not familiar with this kind of stuffs, I'll keep you informed in a second time.

For the tests, agreed, no need of test third party tools, only that our classes behave properly.

I be back when I've some news!

chalasr:
@slashfan 
- Tests added for Encoders (unit+functional, see `Tests\Encoder` & `Tests\Functional\Encoder`).
- Documentation updated:  
  Configuration reference: Add the new options + Mention the encoder customization chapter.  
  Encoder customization: Present the built-in encoders + Give a list of available algorithms.

Let me know about your eventual suggestions (especially about naming and documentation).
Could you setup Scrutinizer as PR-check for the 2.0 branch? It's useful to catch possible optimisations and typos.

Have a good review :)

Resources/config/services.xml
```diff
@@ -19,9 +21,15 @@
     </parameters>
```
weaverryan: For a 2.0 release, the class parameters could be removed. But let's do it in a different PR (later) to keep things focused here :)

weaverryan:
Hi guys!

Really happy to see this stuff :). I started reviewing and I have one suggestion I'd like you to consider: *avoid* renaming services or classes (unless we have a real reason). For example, the `OpenSSLKeyLoader` class moved to a new namespace and its service name was changed slightly. If we can minimize these changes, we could make it much easier for people to upgrade. We *could* even considering making changes in a BC way like the Symfony core does, though this is a bit more work to accomplish :).

Let me know what you think. And thank you!

Resources/config/services.xml
chalasr: @weaverryan Totally agreed. As long as I worked on different files I saw things that need to be changed, such as the documentation that needs more clean examples with namespaces, following best practices. It itches me to change them, but as you said, it's out of the scope of this PR. So, I'll propose this changes in a next one. Thank you for the review!

chalasr:
@weaverryan 
I don't considered this before so thank's for telling me.
Do you think that, because the bundle is based on the Namshi/JOSE library by default, it could be clean to don't differentiate it by its namespace, and so remove the `NamshiJOSE\` part?
The goal is to provide more encoders based on different libraries in the future, and so differentiate them like this.
The main difficulty is that we now have the choice between `OpenSSL` and `SecLib`, and so it would be strange to have one `JWTEncoder` (old, instead of new `OpenSSLEncoder`) and one `SecLibEncoder`, right?
But, as the two classes represent ATM only a constructor (all the logic is inside the `BaseEncoder`), with a bit of imagination I can quickly remove them in order to use the BaseEncoder as the only class (can be keep its old name, `JWTEncoder`), have only one service, and play with arguments to make the service instantiates the class with the good ones, depending on configuration. 
Like this, we could keep services and classes names pretty much same as before. 
It would be nice to have your opinion.

Also, what do you mean by "making changes in a BC way like the Symfony core does"?
Keep the old things while adding new ones and mark them as deprecated before remove them in a next release?

I have to learn a lot about keeping BC as possible and letting the time to upgrade, I'm not sure to do the good choices, So many thanks for your help on these several questions.

@slashfan What do you think about this?

weaverryan:
> Do you think that, because the bundle is based on the Namshi/JOSE library by default, it could be clean to don't differentiate it by its namespace, and so remove the NamshiJOSE\ part?

If the library were new today, then I think having the `NamshiJOSE` makes sense. But I don't think adding it adds enough value for the added BC breaking. For that small reason, I would prefer to not have it.

> The main difficulty is that we now have the choice between OpenSSL and SecLib, and so it would be strange to have one JWTEncoder (old, instead of new OpenSSLEncoder) and one SecLibEncoder, right?

This change - unlike moving the `OpenSSLKeyLoader` class - has a real reason behind it, so I think it makes sense. But, see my comment below about the Symfony way of doing deprecations.

> Like this, we could keep services and classes names pretty much same as before. 
It would be nice to have your opinion.

If you can make everything work with one class, and it's a nice solution (I mean: you're not forcing and hacking things together to make it work), that would be awesome! :)

> Also, what do you mean by "making changes in a BC way like the Symfony core does"?
Keep the old things while adding new ones and mark them as deprecated before remove them in a next release?

Yes, you have exactly the idea :). If you kept the old way of doing things and just deprecated them, then we could merge this into the 1.x branch (as you wouldn't be breaking BC). If you followed the Symfony philosophy entirely, when 2.0 is released, we would simply remove the old deprecated features (this was done with Symfony 2.8 and 3.0). To give a specific example, if you *did* rename the `JWTEncoder` class, you could actually keep it, but make the new class extend the deprecated class, and trigger a deprecation warning in the old class. These things are tricky to get right (even for me) - I often look in Symfony's core for the different strategies on how to properly deprecate things. And honestly, in a smaller library like this (small compared to Symfony), we need to be pragmatic: doing perfect deprecations can take some time. But if you're interested in learning about it, this is a great opportunity.

Cheers!

chalasr:
@weaverryan Thank you very much for take the time to educate me on this way, it will be helpful for my next contributions on Symfony and community bundles.

Create the two services is the "easy" way, because the logic is already in place , we can use the `encoder_service` configuration option to choose one of the available services, rather than creating a new `encoder_engine` option that could be useless if the bundle is used with an other library than NamshiJOSE.
However, I can't bring all the features added here inside the old `JWTEncoder` so, as you said, I'll deprecate it and keep its class+service intact, in order to remove it in a next release.
Also I'll remove the `NamshiJOSE` part of the new encoders namespace, and bring them into only one class (the BaseEncoder, I need help to name this new class, as it will be located in the same directory of the deprecated one, maybe `DefaultEncoder`?), then use it as `class` for the two new services (contrarily to the classes, I can't bring them together in one because their existence makes that's easy to switch).

Please let me know if I'm wrong, if I misunderstood or if you see a better way for something :)

I'll wait for the opinion of @slashfan and do my best for make upgrading as soft as possible for developers, while providing these new features.

chalasr:
Encoder update:

- Moved the `Encoder\NamshiJOSE\BaseEncoder` to `Encoder\DefaultEncoder`
- Re-added the old JWTEncoder class+service and deprecate them since 2.0 (to be removed in ?? ATM I said 2.1)
- Made the new `Encoder\DefaultEncoder` extends the actual (old) `JWTEncoder`.
- Removed the `NamshiJOSE\OpenSSLEncoder` + `NamshiJOSE\SecLibEncoder` and made the two corresponding services use the `DefaultEncoder` class, and so the `NamshiJOSE` namespace has been totally removed as its our default implementation from the beginning.
- Changed the two new service names from `lexik_jwt_authentication.encoder.namshi_jose.[engine]` to `lexik_jwt_authentication.encoder.default.[engine]`

About the `OpenSSLKeyLoader`, I tried to follow the @weaverryan advices as possible and so marked it as deprecated while keeping the class and service, that are used in the deprecated JWTEncoder.
But, because of the new support of SecLib, I cannot avoid the new structure (`Services\KeyLoader` namespace) and make the new `OpenSSLKeyLoader` extending the old, because the new already extends the `AbstractKeyLoader`.
I think this old `OpenSSLKeyLoader` can be quickly removed and inject the new `OpenSSLKeyLoader` in the deprecated `JWTEncoder`.

I think that keep the `JWTEncoder` is

Resources/config/services.xml
```diff
@@ -43,10 +57,22 @@
...
+            <argument>%lexik_jwt_authentication.pass_phrase%</argument>
+        </service>
+        <!-- SecLib Key Loader -->
+        <service id="lexik_jwt_authentication.key_loader.seclib" class="%lexik_jwt_authentication.key_loader.seclib.class%">
+            <argument>%lexik_jwt_authentication.private_key_path%</argument>
+            <argument>%lexik_jwt_authentication.public_key_path%</argument>
+            <argument>%lexik_jwt_authentication.pass_phrase%</argument>
+        </service>
+        <!--- OpenSSL Key Loader / Old implementation (deprecated since 2.0) -->
+        <service id="lexik_jwt_authentication.openssl_key_loader" class="Lexik\Bundle\JWTAuthenticationBundle\Services\OpenSSLKeyLoader">
```
weaverryan: You can officially deprecate these - which will cause the user to see a deprecation warning :) http://symfony.com/blog/new-in-symfony-2-8-deprecated-service-definitions

Resources/config/services.xml
```diff
@@ -6,10 +6,12 @@
...
+        <parameter key="lexik_jwt_authentication.encoder.default.class">Lexik\Bundle\JWTAuthenticationBundle\Encoder\DefaultEncoder</parameter>
         <parameter key="lexik_jwt_authentication.jwt_manager.class">Lexik\Bundle\JWTAuthenticationBundle\Services\JWTManager</parameter>
         <parameter key="lexik_jwt_authentication.handler.authentication_success.class">Lexik\Bundle\JWTAuthenticationBundle\Security\Http\Authentication\AuthenticationSuccessHandler</parameter>
         <parameter key="lexik_jwt_authentication.handler.authentication_failure.class">Lexik\Bundle\JWTAuthenticationBundle\Security\Http\Authentication\AuthenticationFailureHandler</parameter>
-        <parameter key="lexik_jwt_authentication.openssl_key_loader.class">Lexik\Bundle\JWTAuthenticationBundle\Services\OpenSSLKeyLoader</parameter>
+        <parameter key="lexik_jwt_authentication.key_loader.openssl.class">Lexik\Bundle\JWTAuthenticationBundle\Services\KeyLoader\OpenSSLKeyLoader</parameter>
+        <parameter key="lexik_jwt_authentication.key_loader.seclib.class">Lexik\Bundle\JWTAuthenticationBundle\Services\KeyLoader\SecLibKeyLoader</parameter>
```
weaverryan: I think we can avoid adding *new* parameters at least (but as we discussed, remove the old ones in a different PR)

Encoder/JWTEncoder.php
```diff
@@ -26,6 +27,10 @@ class JWTEncoder implements JWTEncoderInterface
     public function __construct(OpenSSLKeyLoader $keyLoader)
     {
         $this->keyLoader = $keyLoader;
+
+        @trigger_error(
+            sprintf('The %s class is deprecated since version 2.0 and will be removed in 3.0, use Lexik\Bundle\LexikJWTAuthenticationBundle\Encoder\DefaultEncoder instead.', get_class($this)
+        ), E_USER_DEPRECATED);
```
weaverryan: nice!

Encoder/DefaultEncoder.php
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @throws \RuntimeException If the JWS cannot be signed
+     */
+    public function encode(array $data)
+    {
+        $jws = new JWS(['alg' => $this->encryptionAlgorithm], $this->encryptionEngine);
```
weaverryan: What do you think about actually configuring the `JWS` instance as a provider service and then injecting that directly into this class? If we added an additional configuration param - `encoder_algorithm` - that defaults to `OpenSSL`, we would only need to register *one* `JWS` service and *one* encoder service (instead of 2). Is there any reason why someone would want to have two services available at once (`OpenSSL` AND `SecLib`)? If not, then one service would be better.

Encoder/DefaultEncoder.php
```diff
@@ -0,0 +1,163 @@
...
+        $key = $this->keyLoader->loadKey('private');
+
+        $jws->setPayload($data);
+
+        try {
+            $jws->sign($key, $this->keyLoader->getPassphrase());
+        } catch (InvalidArgumentException $e) {
+            throw new \RuntimeException(
```
weaverryan: What do you think about creating a custom exception class here? Unfortunately, it looks like Jose doesn't have custom exception classes - they just throw the `\InvalidArgumentException`. I think we should either allow that exception to be thrown, or go the extra step and create a custom exception from our library that they can catch - e.g. `KeySigningException` (again, if we actually knew WHAT went wrong, it would be better to have specific exceptions for the specific problems - but Namshi doesn't allow that as far as I can see). I like your message about "checking your configuration"

Services/KeyLoader/AbstractKeyLoader.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * @param string $type The key type, must be "public" or "private"
+     *
+     * @return string The path of the key
+     *
+     * @throws \InvalidArgumentException If the given type is not valid.
+     */
+    public function getKeyPath($type)
```
weaverryan: It looks like this should really be protected, except that a few tests rely on it. We should at least mark it with an `@internal` - it's not meant to be used from the outside.

Services/KeyLoader/AbstractKeyLoader.php
```diff
@@ -0,0 +1,92 @@
...
+     *
+     * @internal
+     *
+     * @throws \RuntimeException
+     */
+    protected function createUnreadableKeyException($type, $path)
+    {
+        return new \RuntimeException(
+            sprintf('%s key "%s" does not exist or is not readable. Did you correctly set the "lexik_jwt_authentication.%s_key_path" parameter?', ucfirst($type), $path, $type)
+        );
```
weaverryan: nice message!

Encoder/DefaultEncoder.php
weaverryan: Actually, this might not be necessary. The only thing that can go wrong here is if the developer makes a mistake (there's no weird user run-time behavior that we might want to catch). So, the exception class is not important imo.

Encoder/DefaultEncoder.php
```diff
@@ -0,0 +1,163 @@
...
+                return false;
+            }
+        } catch (InvalidArgumentException $e) {
+            throw new \RuntimeException(
+                sprintf('An error occurred while verifying the JWS (%s). Please check your configuration.', $e->getMessage())
+            );
+        }
+
+        return $payload;
+    }
```
weaverryan: There's a slight problem with this method - it exists in the current implementation too. There are 3 big reasons why this method might fail:

a) the token is invalid - it literally can't be decoded
b) the token is expired
c) the token's signature failed

But, there's no way to know which occurred - all return false. What do you think about throwing an exception instead of returning false? We could have 3 exceptions, each which extend the same base exception class (something like `TokenDecodeFailureException`. This is similar to how Guzzle behaves: it throws an exception when a request fails (and has different exception classes for the different failure types)

Encoder/DefaultEncoder.php
```diff
@@ -0,0 +1,163 @@
...
+        return $this->encryptionEngine;
+    }
+
+    /**
+     * @return string
+     */
+    public function getEncryptionAlgorithm()
+    {
+        return $this->encryptionAlgorithm;
+    }
```
weaverryan: We should find a way to accomplish our tests, without adding these getter functions. We need to minimize the public functions of course :)

weaverryan:
Feedback sent - nice job!

Resources/config/services.xml
chalasr: @weaverryan I was glad to discover this feature, but as we have some tests running on old symfony versions, the tests don't pass anymore :/ So I had to rollback. I need to discuss with maintainers to know if we keep the compatibility with symfony versions older than 2.8. What do you think about this?

Encoder/DefaultEncoder.php
chalasr: @weaverryan This is the most difficult decision of this changes, I needed some suggestions to do take the good one, so thank's for your help. 

Having two services is effectively not necessary. If I add an `encoder_engine` config option (The `encoder_algorithm` one is used to choose between the algorithms available for each engine, e.g. RS256, ES256, ...), we could safely have only one `lexik_jwt_authentication.encoder.default`. 
I'll go in this way (add this additional `encoder_engine` option, default to "OpenSSL"), custom encoders can use it or not, such as done for the `encoder_algorithm`.
Then, I just have to replace the two services by one and inject the value of the `encoder_engine` option as the second argument (`$encryptionEngine`) of the service.

About use the JWS as a provider and inject it in the encoder service, I'm not sure that's necessary because deal with the `encoder_engine` should be sufficient, as pretty much all can be done with the JWS class in the new DefaultEncoder (it just takes the encryptionEngine as constructor argument, it's already in place and managed with the two different services because of the missing `encoder_engine` option).
I propose you to add the  `encoder_engine` option and be back, then if you think that could make sense to do it, you could talk me more about this way sort as I can try to implement it.

Encoder/DefaultEncoder.php
chalasr: I tried to test most of possible use cases, the exception thrown by namshi/jose say "Invalid key supplied" if it can't use the given key with its passphrase, so I added this try-catch to give an indication about what is wrong and from where the problem come from, I also give the message of the previous exception in the message of the last thrown exception.
Do you think it makes sense to have this catch?

Services/KeyLoader/AbstractKeyLoader.php
chalasr: Good catch! 👍

Encoder/DefaultEncoder.php
chalasr: Agreed. The return of this method is handled in the [JWTProvider](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Security/Authentication/Provider/JWTProvider.php#L64), it throw the same AuthenticationException for the 3 possible cases you given ("Invalid token"). We can easily give a more precise message by creating these 3 exceptions and separate the different checks that would throw the corresponding exception.
But, as the JWTProvider needs to handle these exceptions properly (throw the same AuthenticationException but by catching the new exceptions and replace "invalid token" by the exception message for instance, instead of just wait for a bool), don't you think that this should be done in a next PR? Sort as this one doesn't touch the JWTProvider at all.

Resources/config/services.xml
weaverryan: Ah, that's interesting! As long as the class itself is deprecated (with the trigger), it's not big deal.

Encoder/DefaultEncoder.php
weaverryan: I *think* I like the catch, because you used it nicely to add more info to the message

Encoder/DefaultEncoder.php
weaverryan: Hmm, so if we make this change (which I think we should), we're actually slightly changing the interface of both `JWTEncoderInterface::decode` and `JWTManagerInterface::decode()` - as these *appear* to not advertise that an exception will ever be thrown. Because you kept BC so well with the original `JWTEncoder`, you were able to not update `JTWProvider` at all, which is pretty cool :).

So we need to be pragmatic to not make too many breaks... but make them when it has benefit. Technically, in the Symfony core, we'd probably make 2 new interfaces and deprecate the old ones... but that's overkill. I think we should:

A) Update the phpdoc of the interfaces to say that a certain type(s) of exception will be thrown. Make this as a BC break - perhaps add a new CHANGELOG.md file to the repo (a good idea anyways, so we can document the upgrade path).
B) Update (in this PR) this class to throw the exceptions
C) Then obviously, update JWTProvider

We could do this in 2 PR's, but it seems quite related, so let's get it done on one spot I think.

slashfan:
Nice work guys ! Sorry I don't have time to help more right now, but thanks for the efforts ! Let us know when the PR is OK for you.

slashfan:
@weaverryan Great idea for the CHANGELOG.md !

chalasr:
@weaverryan @slashfan Ready for review. 
A new BC break added in the configuration by making `lexik_jwt_authentication.encoder` an ArrayNode, I think it really makes sense for a matter of structure. 
Also, standards about Exception structure and naming are very rare, I tried to stay logic without having too long FQCNs.
Thank's by advance.

slashfan:
OK for me, I just merged #164

chalasr:
@slashfan Could you merge #164 in 2.0 sort as I can rebase this one on top of 2.0 and update the changelog? Stop me if I'm wrong. Thank's

slashfan:
@chalasr No problem, it's done

weaverryan:
I'll properly review and test this soon :)

chalasr:
@weaverryan You already spent a considerable time here, thank's for that.
As this PR touch several places, I see changes that particularly needs your eye:
- JWSProvider (thank's to your suggestion)
- Exceptions __naming__ and directory __structure__
- Services __accessibility level__
- Configuration
- Documentation

Waiting for your feedback :)

Encoder/DefaultEncoder.php
```diff
@@ -0,0 +1,80 @@
...
+     */
+    public function encode(array $data)
+    {
+        try {
+            $this->jwsProvider->createSignedToken($data);
+        } catch (InvalidArgumentException $e) {
+            throw new JWTEncodeFailureException('An error occurred while trying to encode the JWT token.', $e);
+        }
+
+        if (JWSProviderInterface::SIGNED !== $this->jwsProvider->getStatus()) {
```
weaverryan: The `getStatus()` method isn't part of the `JWSProviderInterface`. And actually, I think we should do this a different way. Having `status` on the `JWSProvider` gives the service "state" - weird things could happen if you were encoding or decoding multiple tokens in a request (I know, edge case). The state should be (somehow) either returned from the `createdSignedToken` and `loadSignature` or an exception should be thrown in this case. But I don't think we should store the status on the service itself.

Exception/JWTDecodeFailure/UnverifiedJWTDecodeFailureException.php
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @author Robin Chalas <robin.chalas@gmail.com>
+ */
+class UnverifiedJWTDecodeFailureException extends JWTDecodeFailureException
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct(
+        $message = 'Unable to verify the given JWT through the given configuration. If the "lexik_jwt_authentication.encoder" encrytion options have been changed since your last authentication, please renew the token. If the problem persists, verify that the configured keys/passphrase are valid.',
```
weaverryan: Typo :) `encrytion`

Security/Authentication/Provider/JWTProvider.php
```diff
@@ -117,4 +122,16 @@ public function setUserIdentityField($userIdentityField)
     {
         $this->userIdentityField = $userIdentityField;
     }
+
+    /**
+     * @param JWTDecodeFailureException $previous
+     *
+     * @return AuthenticationException
+     */
+    private function createInvalidJWTException(JWTDecodeFailureException $previous = null, $message = 'Invalid JWT Token')
```
weaverryan: Small thing, but we *never* pass the `$message` arg - so we could just *not* have this be an argument at all.

Services/JWSProviderInterface.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace Lexik\Bundle\JWTAuthenticationBundle\Services;
+
+/**
+ * Interface for JWS provider classes.
```
weaverryan: Interface for classes that are able to create and load JSON web signatures (JWS).

Services/JWSProviderInterface.php
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @since 2.0
```
weaverryan: It's up to you, but I don't see the `@since` very often - with git/GitHub, it's so easy to find this information, if you actually need it (and then you don't have to keep it up to date manually)

Resources/config/services.xml
```diff
@@ -19,9 +19,14 @@
...
     <services>
-        <!-- JWT Encoder / Decoder / Default implementation -->
+        <!--  Default JWT Encoder / Decoder -->
+        <service id="lexik_jwt_authentication.encoder.default" class="Lexik\Bundle\JWTAuthenticationBundle\Encoder\DefaultEncoder">
+            <argument type="service" id="lexik_jwt_authentication.jws_provider"/>
+        </service>
+        <!-- JWT Encoder / Decoder / Old implementation (deprecated) -->
         <service id="lexik_jwt_authentication.jwt_encoder" class="%lexik_jwt_authentication.jwt_encoder.class%">
             <argument type="service" id="lexik_jwt_authentication.openssl_key_loader"/>
+            <deprecated>The "%service_id%" is deprecated since version 2.0 and will be removed in 3.0, use "lexik_jwt_authentication.encoder.default" instead.</deprecated>
```
weaverryan: Now that I am playing with your updated code, I think I'm going to change my mind :). I think we should remove the 2 deprecated services (`lexik_jwt_authentication.jwt_encoder` & `lexik_jwt_authentication.openssl_key_loader` and their classes. We can simply add this to an `UPGRADE` file. In both cases, the new service works *almost* identically to the old. And unless you're working with these services in some deep way, it will just work.

For me, in the KnpU tutorial, I was able to upgrade the library, change 2 spots to reference `lexik_jwt_authentication.encoder` instead of `lexik_jwt_authentication.jwt_encoder` (which, probably I should have been using anyways), and everything just worked. Even my old JWT's worked out-of-the-box with the new system. Nice job :)

Encoder/DefaultEncoder.php
```diff
@@ -0,0 +1,80 @@
...
+
+        return $this->jwsProvider->getToken();
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @throws JWTDecodeFailureException           If the signature cannot be loaded
+     * @throws UnverifiedJWTDecodeFailureException If the signature cannot be verified
+     * @throws ExpiredJWTDecodeFailureException    If the token is expired
```
weaverryan: All of the details about the exceptions should be moved to the interface. This is one of the "BC" breaks we need to add to an UPGRADE file. Specifically, both `encode()` and `decode()` previously did *not* throw exceptions on failures, but now they do. We "changed" how the interface behaves :) (it's a good change!). But if you're upgrading, you'll need to know about this. For example, in my KnpU tutorial, we are checking to see if the result of decode is `false` - now we'll need to catch the exceptions.

Resources/doc/5-encoder-service.md
```diff
@@ -4,7 +4,7 @@ JWT encoder service customization
 This bundle comes with a [`namshi/jose`](https://github.com/namshi/jose) library based token encoder as it uses SSH keys to encrypt and decrypt data.
 If this doesn't suit your needs, you can replace it with your own encoder service. Here's an example implementing a [`nixilla/php-jwt`](https://github.com/nixilla/php-jwt) library based encoder.
 
-#### Create the encoder class
+### Create the encoder class
 
 ``` php
 // src/Acme/Bundle/ApiBundle/Encoder/NixillaJWTEncoder.php
```
weaverryan: not part of this PR, but we should update the docs to use `AppBundle` - since it's what people expect :)

Services/JWSProvider.php
```diff
@@ -0,0 +1,192 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function createSignedToken(array $payload)
+    {
+        $this->jws = new JWS(['alg' => $this->encryptionAlgorithm], $this->encryptionEngine);
+        $this->jws->setPayload($payload);
+        $this->sign();
+
+        return $this;
```
weaverryan: I already commented about "state" in this class. So, we need to find a different way to handle this stuff. It looks like you're basing the behavior on the `JWS` class itself, which makes sense - but it's wrong in the same way - it's super-stateful. Ideally, we can "correct" that in our class at least :).

One option is to return a new value object from `createSignedToken` - something with a `getToken()` method and a `getState()` method (or perhaps better, an `isSigned()` method). For `loadSignature()`, you would return another value object with `getStatus()` (or better ,`isVerified()`), `isExpired()` and `getPayload()`. The `JWSProviderInterface` would then only have `createSignedToken()` and `loadSignature()` (but you could re-name these after the update, if there is a better name then).

Services/JWSProvider.php
```diff
@@ -0,0 +1,192 @@
...
+        return $this->jws;
+    }
+
+    /**
+     * @param string $encryptionEngine
+     * @param string $encryptionAlgorithm
+     *
+     * @return bool
+     */
+    protected function isAlgorithmSupportedForEngine($encryptionEngine, $encryptionAlgorithm)
```
weaverryan: Also, I would make *everything* `private`, unless you actively have a reason that you need `protected`. This will make it much easier to keep BC later. If someone needs to override something, they can open an issue and we can change to protected (or, we may be able to tell them the "correct" way to accomplish what they want).

Services/KeyLoader/AbstractKeyLoader.php
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * @var string
+     */
+    protected $publicKey;
+
+    /**
+     * @var string
+     */
+    protected $passphrase;
```
weaverryan: I'd make these all `private` too - they're not needed by the subclasses

Services/KeyLoader/AbstractKeyLoader.php
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * @throws \InvalidArgumentException If the given type is not valid.
+     */
+    protected function getKeyPath($type)
+    {
+        $property = $type . 'Key';
+
+        if (!property_exists(get_class($this), $property)) {
+            throw new \InvalidArgumentException(sprintf('The key type must be "public" or "private", "%s" given.', $type));
+        }
```
weaverryan: This could be less magic:

```php
if ($type === 'public') {
    return $this->publicKey;
}

if ($type === 'private') {
    return $this->privateKey;
}

throw new \InvalidArgumentException('...');
```

Resources/config/services.xml
```diff
@@ -44,10 +49,23 @@
...
+            <argument>%lexik_jwt_authentication.private_key_path%</argument>
+            <argument>%lexik_jwt_authentication.public_key_path%</argument>
+            <argument>%lexik_jwt_authentication.pass_phrase%</argument>
+        </service>
+        <!-- SecLib Key Loader -->
+        <service id="lexik_jwt_authentication.key_loader.phpseclib" class="Lexik\Bundle\JWTAuthenticationBundle\Services\KeyLoader\SecLibKeyLoader">
+            <argument>%lexik_jwt_authentication.private_key_path%</argument>
+            <argument>%lexik_jwt_authentication.public_key_path%</argument>
+            <argument>%lexik_jwt_authentication.pass_phrase%</argument>
+        </service>
```
weaverryan: I think you can make both of these individual key loaders private. Then, nobody will use them directly - they'll just use the alias.

Resources/config/services.xml
```diff
@@ -19,9 +19,14 @@
     </parameters>
 
     <services>
-        <!-- JWT Encoder / Decoder / Default implementation -->
+        <!--  Default JWT Encoder / Decoder -->
+        <service id="lexik_jwt_authentication.encoder.default" class="Lexik\Bundle\JWTAuthenticationBundle\Encoder\DefaultEncoder">
+            <argument type="service" id="lexik_jwt_authentication.jws_provider"/>
+        </service>
```
weaverryan: Just like with the key loaders, I think you can make this private. Then people will be forced to use the alias `lexik_jwt_authentication.encoder`, which is what we want.

weaverryan:
Sorry for the delay @chalasr!

First, this worked *awesome* when I tried it - so easy moving from the old version - that's huge :).

I left comments, but I'm very happy with this. Here's a summary of the blockers, which I think will be easy to fix:
A) Removing the deprecated classes/services - that's my fault, I completely changed my mind on that.
B) Adding an UPGRADE file. From what I can see, it should include the re-naming of the command (`lexik:jwt:check-open-ssl`), details about how the `JWTEncoderInterface` methods now throw exceptions, details about the 2 service ids that were removed (from A) and details about the 2 classes that were removed/moved (the classes from A).
C) Removing the "statefullness" of JWSProvider / JWSProviderInterface. 

And, that's it :). The configuration looks good, exception classes look good, and we'll save documentation for a different PR - you updated the most important parts (configuration). For normal users, the changes in this PR will be invisible.

Cheers!

Resources/config/services.xml
chalasr: You just made my day :).

chalasr:
Glad to hear that "upgrading" was not too hard. 

I was very not sure about the JWSProvider state, it will not be hard to find another way from your suggestions.

It's ok for the rest, I was looking for these notes.
I'll ping you when changes are done for a last eye.

Thank's for your nice review @weaverryan

weaverryan:
Cheers! Looking forward to seeing it after the changes - I think it will be a fast & easy merge :)

Services/JWSProvider.php
chalasr: 👍  for the value object, a great input! 

I ran out of inspiration to find a good location for this one. Maybe you can help me?
My first thought is to put it inside the Encoder namespace that would give `Encoder/Token.php`.
Do you think that it makes sense?
Maybe it should have its own namespace+directory...
Thank's in advance

weaverryan: Hmm, I'm not sure either :). If you created a new directory called `Signature` and moved the `JWS` stuff there, then you could create the model classes there - e.g .`Signature\Signature` (that one looks a little weird, but maybe this is good inspiration at least)

chalasr: I'll start from this, maybe you'll be able to help me more by seeing it in action.
Thank you for the try :)

Resources/doc/5-encoder-service.md
chalasr: + add namespace to controllers as `AppBundle\Controller` for all code examples 👍

chalasr:
@weaverryan I made the changes, it looks awesome to me!

All is more decoupled, each class has its own role and can be tested as well (and it is!). 
I named the two model classes `CreatedJWS` and `LoadedJWS` to make the separation clear, I would like to have your opinion.

@slashfan I need your help and more, I introduced a new directory called `Signature` that contains two models and should not evolve ATM. I have a doubt about moving it inside the Encoder namespace but not sure it sounds good. Please let me know about your suggestions, also if you can give me your opinion about naming of these classes, I'm totally open for better ones.

Hopefully waiting for your feedback guys :)

chalasr:
ping @slashfan @weaverryan :blush:

slashfan:
@chalasr :+1: I have no problem with the Signature directory being outside of the Encoder directory. 

Just a thought, `Token/Signature` directory organization might be nice too. Along with `Token/Encoder` and `Token/Extractor` directories... but it might be too much work and bc breaks.

chalasr:
`Token/Signature` sounds also good to me, this new directory is a bit hard to place so there is no ideal solution. I'll do the change depending on @weaverryan opinion for make it comprehensive.

chalasr:
@weaverryan Your opinion is really important as you given the way, but now I think it's time to merge this, in order to go with next steps on v2 and see a first relevant change for.

I be back in 2-3 days, I hope I'll find some news from you about https://github.com/lexik/LexikJWTAuthenticationBundle/pull/162/commits/b480ed623b4059b8afb442454a0b79180295b6d7, otherwise it doesn't matter, your feedbacks wouldn't be lost.

PR #167, opened 2016-05-10
Remove Symfony versions < 2.8 from travis.yml

chalasr:
After discussed in #132, we stop supporting Symfony versions lower than 2.8 for the branch 2.0 and `2.x` releases of this bundle, and so stop ensuring compatibility with these Symfony versions on travis-ci .

Open to any suggestion (missing version, specific php version for specific symfony one, ...).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,16 +10,6 @@ php:
 matrix:
   fast_finish: true
   include:
-    - php: 5.4
-      env: SYMFONY_VERSION=2.3.*
-    - php: 5.5
-      env: SYMFONY_VERSION=2.3.*
-    - php: 5.6
-      env: SYMFONY_VERSION=2.3.*
-    - php: 5.6
-      env: SYMFONY_VERSION=2.7.*
-    - php: 7.0
-      env: SYMFONY_VERSION=2.7.*
     - php: 5.6
       env: SYMFONY_VERSION=2.8.*
     - php: 7.0


slashfan:
We should remove PHP5.4 reference and also test against PHP 5.5, I know it's dead but this is Symfony 3.0 and current namshi/jose minimal PHP version requirement.

chalasr:
Don't you think that we should ensure compatibility with (almost) the minimal PHP version required by Symfony 2,8? It requires 5.3.9 so 5.4 could be a good compromise, plus the 5.4 release given lots of new features.

Otherwise no problem, I'll remove it.

slashfan:
OK, we can keep PHP 5.4 compatibility since we do not directly use specific PHP 5.5+ features at the moment (athough namshi/jose library currently requires 5.5+ so there might be a conflict during a composer install). But PHP 5.3 is way too old to be taken care of anymore.

chalasr:
@slashfan Did I miss something or is it ok for this one?

slashfan:
@chalasr OK for me

chalasr:
@slashfan You were right, 5.5 introduced interesting features such as `ClassName::class` to get a FQCN.
I'll remove 5.4 form travis config and make the bundle requiring php 5.5.
Is it ok for you?

slashfan:
:+1: for me

PR #170, opened 2016-05-14
PSR-4 Autoloading

chalasr:
> The PSR-4 autoload standard is recommended for modern bundles

From [Symfony best practices for reusable bundles](http://symfony.com/doc/current/cookbook/bundles/best_practices.html#composer-metadata).


slashfan:
:+1:

PR #174, opened 2016-05-19
Patch travis versions

chalasr:
As discussed in #167, we stop testing on PHP versions < 5.5.


chalasr:
bad branch ... Reopened in #175

PR #176, opened 2016-05-20
Compatibility php 5.3

fconte:
A few small changes to make compatible code with PHP 5.3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ build
 vendor
 composer.lock
 composer.phar
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@ build
 vendor
 composer.lock
 composer.phar
-.idea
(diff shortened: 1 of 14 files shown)


chalasr:
Hello @fconte,

Thank's for your work.
Even it could be easy to make the changes, I'm not sure that is a direction where this bundle wants to go.

@slashfan What do you think?

chalasr:
@fconte Depending on the answer of slashfan, could you please add `5.3` to the .travis.yml versions in order to ensure that tests are green?

tristanbes:
PHP `5.3` *and* `5.4` are EOL (end of life).

http://php.net/supported-versions.php

To me it's not the right move to add compatibility to those versions.

chalasr:
@tristanbes Totally agreed.

@fconte Periods between new PHP releases are really shorter than 2 or 3 years ago, upgrade your server version must be too. It's time to move ...

I close this issue.

fconte:
The [LTS version of Symfony supports php 5.3.9,](http://symfony.com/doc/2.8/reference/requirements.html) but I understand and agree that this is not the future.

slashfan:
:+1: @chalasr PHP 5.3 and 5.4 are dead plus the library we use is compatible only php 5.5+.

PR #177, opened 2016-05-20
Add JWTAuthenticationResponse

chalasr:
| Q             | A    |
|---------------|------|
| Bug fix?      | no  |
| New feature?  | yes |
| BC breaks?    | no  |
| Deprecations | yes |
| Fixed tickets | n/a |
| Tests pass?   | yes  |

This introduces a generic Response class for successful/failed authentication.
BTW remove duplicated failure Responses with pretty much the same data (3times).

Let me know what do you think.

Steps:

- [x] Add JWTAuthenticationFailureResponse & JWTAuthenticationSuccessResponses
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/Response/JWTAuthenticationFailureResponse.php
+++ b/Response/JWTAuthenticationFailureResponse.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace Lexik\Bundle\JWTAuthenticationBundle\Response;
+
+use Symfony\Component\HttpFoundation\JsonResponse;
+
+/**
+ * JWTAuthenticationFailureResponse.
+ *
+ * Response sent on failed JWT authentication (can be replaced by a custom Response).
+ *
+ * @internal
+ *
+ * @author Robin Chalas <robin.chalas@gmail.com>
+ */
+final class JWTAuthenticationFailureResponse extends JsonResponse
+{
+    /**
+     * The response message.
+     *
+     * @var string
+     */
+    private $message;
+
+    /**
+     * @param string $message A failure message passed in the response body
+     */
+    public function __construct($message = 'Bad credentials')
+    {
+        $this->message = $message;
+
+        parent::__construct(null, self::HTTP_UNAUTHORIZED, ['WWW-Authenticate' => 'Bearer']);
+
+        $this->setData([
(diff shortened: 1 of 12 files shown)


chalasr:
ping @slashfan

slashfan:
OK for me, I was just waiting for the last checkbox to be checked

chalasr:
I was waiting your opinion about the feature before writing the doc :) I ping you when it's done

chalasr:
@slashfan Doc slightly updated, right to review. No changed typehint and no difference in events that developers listen on, no BC break.

Response/JWTAuthenticationFailureResponse.php
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Sets the failure message.
+     *
+     * @param string $message
+     *
+     * @return JWTAuthenticationFailureResponse
+     */
+    public function setMessage($message)
+    {
+        $this->message = $message;
```
GromNaN: Is it expected that the JSON body keep the initial message when the message is modified ?

Response/JWTAuthenticationSuccessResponse.php
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @param array $data Unused parameter for risks limitation
+     *
+     * @internal
+     */
+    public function setData($data = [])
```
GromNaN: Set data is completely ignored.
The `extraData` property could be removed.

```

public function __construct($token, $data = null, $status = 200, array $headers = array())
{
    $this->token = $token;
    parent::__construct($data, $status, $headers);
}

public function setData($data) {
    parent::setData(['token' => $this->token] + $data);
}
```

chalasr: Wow... I was gone to far, written too late in the night... This will be widely sufficient.

Response/JWTAuthenticationFailureResponse.php
chalasr: Good catch @GromNaN

slashfan:
:+1:

PR #179, opened 2016-06-04
[Proposal] Dev kit

chalasr:
Hi there,

We have some requirements about coding style. They are not very specific in comparison with other bundles, but they are actually not totally applied.

What I propose is to add a `.php_cs` containing all our rules.
This one would be automatically detected when running `php-cs-fixer fix`.

Also I added `friendsofphp/php-cs-fixer` as dev dependency to make contributors able to use it without install it globally.

To make this properly applied, I propose to add StyleCI as PR check.
StyleCI is fast to run and easy to use (it gives patches easy to apply by copy/paste or real `git checkout`).

In addition of this, I ran a `php-cs-fixer fix` with the configuration on our current code.
There is some fixes that I added in a specific commit on this PR.
Let me know if you think that it should not be merged, I'd remove it. But I think it's quite better to let contributors start with a clean code base, without caring of which file they have to fix then add to their commits (stay in the PR scope).

Please let me know what you think.
Also contributors opinions would be really appreciated, as this should make contributions easier and not the inverse.
Thanks

-----------

- [x] Add `.php_cs` config file, default to `symfony` level with custom fixers added and removed
- [x] Require `friendsofphp/php-cs-fixer` as dev dependency
- [x] Add StyleCI as PR check (@slahsfan I need that you setting it up, if it sounds good to you of course)
- [x] Document the `php-cs-fixer` usage in CONTRIBUTING.md

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -7,7 +7,6 @@ $finder = Symfony\CS\Finder\DefaultFinder::create()
 return Symfony\CS\Config\Config::create()
     ->level(Symfony\CS\FixerInterface::SYMFONY_LEVEL)
     ->fixers(array(
-        '-phpdoc_short_description',
         '-phpdoc_scalar',
         '-extra_empty_lines',
         'single_blank_line_before_namespace',
--- a/.php_cs
+++ b/.php_cs
@@ -9,9 +9,7 @@ return Symfony\CS\Config\Config::create()
     ->fixers(array(
         '-phpdoc_scalar',
         '-extra_empty_lines',
-        'single_blank_line_before_namespace',
-        'unalign_double_arrow',
-        'unalign_equals',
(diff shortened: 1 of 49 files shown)


slashfan:
:+1: Good idea @chalasr. I just added the repo to StyleCi (https://styleci.io/repos/18628501).

chalasr:
@slashfan Great! Please let me know if you see any missing [fixer](https://github.com/friendsofphp/php-cs-fixer#usage) in the [config file added](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/179/files#diff-b31b99fa4c2c185ba6a71235efb48224).

slashfan:
👍 @chalasr perfect for me !

PR #184, opened 2016-06-16
[Security] Deprecate current system in favor of a JWTTokenAuthenticator (Guard)

chalasr:
| Q             | A    |
|---------------|------|
| Bug fix?      | no  |
| New feature?  | yes |
| BC breaks?    | no  |
| Deprecations | yes |
| Fixed tickets | ~ |
| Tests pass?   | yes  |

See #132 discussion.

@slashfan after you started to work on, I made some changes.
I'm really not an expert in the Guard component itself, but it seems make our life easier from all sides.

~~Should we propose it as an alternative or deprecate the "legacy" way in favor of this?~~

~~What's next?~~

---

- [x] Add the Guard JWTAuthenticator
- [x] Test it
- [x] Add a full functional test case
- [x] Depreciate old security system
- [x] Exclude legacy tests from CI
- [x] Update the documentation


chalasr:
@weaverryan Your help would be really appreciated, as your are the father of this great tool (guard).

slashfan:
@chalasr The main idea for V2 is to replace the legacy system by a new one based on the Guard Component. It would help transition but it would also mean too much work to maintain both systems.

chalasr:
@slashfan Agreed, though upgrading can stay soft using the JWTGuardFactory that could reproduce the same configuration as for the legacy system (`lexik_jwt: ~`).
I would just change the return of `getKey()` to `lexik_jwt` rather than `lexik_jwt_guard`, sort as we don't have any breaking changes in security configuration, expect if people make the change themselves (as documented). What do you think?

We also even need some security options such as `throw_exceptions`, `create_entry_point` and the choice of token extractors, right? They could be kept through this way.

slashfan:
@chalasr Agreed ! Let's try to keep the transition smooth, we'll remove the old system in a V3

chalasr:
@slashfan I'm facing an issue.

One of the main advantages of Guard is that the end user can have several authenticators per firewall.
The problem is: how to apply our current security configuration on the new guard authenticator?
It can be done using the factory I added (an override of the original GuardAuthenticationFactory that's not the normal usage of this component) that would be used in the same way as our current system:

```yaml
firewalls:
    api:
        lexik_jwt_guard: ~
```

Instead of the "normal" way:

```yaml
firewalls:
    api: 
        guard:
            lexik_jwt_authenticator.jwt_authenticator
```

Because like this (second way), we can't pass any config option. 
I'm afraid that Guard cannot fit our needs because of this. 

I'll investigate, then keep you informed.
Please let me know if you have any input.

chalasr:
See https://github.com/symfony/symfony/issues/19175

slashfan:
We could move the security configuration to the bundle configuration (extractors, prefix) and use it to configure the lexik_jwt_authenticator.jwt_authenticator service. 

By default the bundle would provide one preconfigured authenticator (this one) and if another firewall needs to use JWT with a different configuration the developer would just have to create another authenticator service extending this one and overriding its configuration. 

What do you think ? That would simplify the factory.

chalasr:
@slashfan It seems to be the better alternative, good idea!
The authenticator service would take all our current per-firewall configuration.
I'll try to keep the configuration simple so the service is easily extensible, then document it as well as possible.

Now that we go in this way, I'm not sure about the usefulness of the factory added here, I'm afraid that it could be more confusing than useful, let me know what do you think.

BTW, the branch of this PR is part of the main repo, not on my fork, do not hesitate to make changes as you want.

slashfan:
Even better if we can get rid of the new factory class !

chalasr:
I would like to discuss about some of our current config options, in order to keep them or not during the migration to Guard.

The primary point of the  `throw_exceptions` option is that an end user setting the option to `true` is able to catch it (right?). IMO the authentication success/failure listeners are here for that. Plus, now that we are going to use Guard, an exception can be thrown in `getUser` (as documented in the [GuardAuthenticatorInterface](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Security/Guard/GuardAuthenticatorInterface.php#L73)) causing the authentication fail and returning the common 401 (json) response. 
I think we can remove it, then if an user need this feature, we could explain how to get around the need.

Next, the `create_entry_point` option. IMO this option is no more useful (now that we are going to use Guard too), `guard` has its own `entry_point` option, taking the service id of one of the configured authenticators that should be used as entry point (`GuardAuthenticator::start()`). Of course, we could even keep it and, if set to `false`, make `JWTTokenAuthenticator::start()` not returning a failure response, but it is IMO the exact role of the `anonymous` firewall option, letting the request continue if anonymous requests are allowed, otherwise call the entry point.

That's all.

All opinions are really welcome.
@slashfan What do you think?

TokenExtractor/ChainTokenExtractor.php
```diff
@@ -0,0 +1,54 @@
...
+     */
+    public function extract(Request $request)
+    {
+        foreach ($this->extractors as $extractor) {
+            if ($token = $extractor->extract($request)) {
+                return $token;
+            }
+        }
+
+        return false;
```
chalasr: to be removed, use `string|null` as return in interface instead, then check for `null === TokenExtractorInterface::extract($request)` in authenticator.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ For a diff between two versions https://github.com/lexik/LexikJWTAuthenticationB
 
 ## [2.0](https://github.com/lexik/LexikJWTAuthenticationBundle/tree/2.0)
 
+* feature [\#184](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/184) [Security] Depreciate current system in favor of a JWTTokenAuthenticator (Guard) ([chalasr](https://github.com/chalasr))
```
weaverryan: Deprecate - depreciate I believe has a different (but similar) meaning :)

Resources/doc/1-configuration-reference.md
```diff
@@ -54,44 +70,22 @@ __Supported algorithms for OpenSSL:__
...
-            create_entry_point:      true      # When no authentication details are provided, create a default entry point that returns a 401 response
-            authentication_provider: lexik_jwt_authentication.security.authentication.provider
-            authentication_listener: lexik_jwt_authentication.security.authentication.listener
+        guard:
+            authenticators: 
+                - lexik_jwt_authentication.jwt_token_authenticator
 ```
+
+For more details about the `lexik_jwt_authentication.jwt_token_authenticator` service and how to
+customize it, see ["Extending the Guard JWTTokenAuthenticator"](6-extending-jwt-authenticator.md)
```
weaverryan: This is the first time I know of that a bundle is replacing a traditional listener/provider security system with Guard. So, this library is the "guinea pig" for this :). I'm interested in how this documentation will read, but at first, it does seem really promising: instead if needing to understand various config options, you could instead override methods on a class.

Security/Authentication/Token/PreAuthenticationJWTUserToken.php
```diff
@@ -0,0 +1,55 @@
...
+namespace Lexik\Bundle\JWTAuthenticationBundle\Security\Authentication\Token;
+
+use Symfony\Component\Security\Guard\Token\PreAuthenticationGuardToken;
+
+/**
+ * PreAuthenticationJWTUserToken.
+ *
+ * @author Robin Chalas <robin.chalas@gmail.com>
+ */
+final class PreAuthenticationJWTUserToken extends PreAuthenticationGuardToken
```
weaverryan: I like that you've created a nice object that can be returned from getCredentials(), instead of an array. But, I don't think there's any advantage to extending `PreAuthenticationGuardToken` - you can really just have a class that doesn't return *anything*.

Exception/JWTAuthenticationException.php
```diff
@@ -0,0 +1,69 @@
...
+namespace Lexik\Bundle\JWTAuthenticationBundle\Exception;
+
+use Symfony\Component\Security\Core\Exception\AuthenticationException;
+
+/**
+ * Exception to be used during a failed JWT authentication process.
+ *
+ * @author Robin Chalas <robin.chalas@gmail.com>
+ */
+class JWTAuthenticationException extends AuthenticationException
```
weaverryan: Hmm, I'm not sure about this class:

1) This extends `AuthenticationException`... and the message that will be returned to the client comes from the `getMessageKey()` method, which will be `An authentication exception occurred.` for all of these different situations. Don't we want the messages in this class to actually be shown to the client?

2) Since all these different exception "situations" are the same class, it will make it a bit harder for the end-user to potentially take different action based on what went wrong (e.g. if they want to do something different if the token is invalid versus if the token is not found). Could we use different classes?

Security/Guard/JWTTokenAuthenticator.php
```diff
@@ -0,0 +1,192 @@
...
+            $user = $userProvider->loadUserByUsername($identity);
+        } catch (UsernameNotFoundException $e) {
+            throw JWTAuthenticationException::invalidUser($identity, $identityField);
+        }
+
+        $authToken = new JWTUserToken($user->getRoles());
+        $authToken->setUser($user);
+        $authToken->setRawToken($preAuthToken->getCredentials());
+
+        $this->dispatcher->dispatch(Events::JWT_AUTHENTICATED, new JWTAuthenticatedEvent($payload, $authToken));
```
weaverryan: I like the idea of dispatching events in the authenticator as an extension point :)

Security/Guard/JWTTokenAuthenticator.php
```diff
@@ -0,0 +1,192 @@
...
+     * {@inheritdoc}
+     *
+     * @param JWTFailureEventInterface $event An event to be dispatched (default JWTInvalidEvent)
+     */
+    public function onAuthenticationFailure(Request $request, AuthenticationException $authException, JWTFailureEventInterface $event = null)
+    {
+        if (null === $event) {
+            $event = new JWTInvalidEvent($authException, new JWTAuthenticationFailureResponse($authException->getMessage()));
+            $this->dispatcher->dispatch(Events::JWT_INVALID, $event);
+        }
```
weaverryan: I think having the optional `$event` argument is a little weird. I would simply return `$event->getResponse()` from `start()` - I don't see any advantage in calling this method, and passing it a non-null `$event`. It confused me at first :)

weaverryan:
I really like the direction here :)

Security/Authentication/Token/PreAuthenticationJWTUserToken.php
chalasr: I made it extend `PreAuthenticationGuardToken` primarily because [`JWTManager::decode()`](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/guard_jwt_authenticator/Services/JWTManager.php#L83) that is used in [`getUser`](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/guard_jwt_authenticator/Security/Guard/JWTTokenAuthenticator.php#L84) only accepts a`TokenInterface` instance, which makes sense to me, at least it doesn't worth to change it imho. Secondarily for benefiting of the `PreAuthenticationGuardToken::setAuthenticated()` method throwing an exception and the `$authenticated` property set to `false` in `__construct()`. 
This object inherit lots of properties and methods that aren't necessary, I just didn't find a lighter alternative for now :/

Security/Guard/JWTTokenAuthenticator.php
chalasr: It was just for reuse `onAuthenticationFailure` that was a mistake, I changed it.

Exception/JWTAuthenticationException.php
chalasr: I rely on the fact that the dispatched event should be enough to know the failure reason, but for instance we can't differentiate an invalid token from an expired one without checking `AuthenticationException::getMessage()` (see #223).
I'll spit that class in many 👍  each one extending `AuthenticationException` and overriding `getMessageKey()` (right?)

chalasr:
Nice to see you @weaverryan :) Thanks for giving suggestions and your feeling about this.

Exception/JWTAuthenticationException.php
weaverryan: > I'll spit that class in many 👍 each one extending AuthenticationException and overriding getMessageKey()

That's right!

weaverryan:
@chalasr Happy to be here to help - but don't ever let me hold up progress :)

chalasr:
Suggestions applied, thank you again Ryan.

However I have an issue with one of the events we dispatch:
`JWTAuthenticatedEvent` is dispatched immediately after creating the authenticated `TokenInterface` object, allowing the end user to set custom attributes in the token that will be put in the `TokenStorage` using `TokenInterface::setAttribute()`, especially for setting attributes corresponding to claims of the token payload.
 
In Guard, there is a dedicated method called `createAuthenticatedToken()` that is provided by the `AbstractGuardAuthenticator` and called on success. 
The problem is that we can't create and return this object on the fly in this method because we need to access the raw token and its payload for building the `TokenInterface` object (our custom `JWTUserToken`). Actually it must be done in `getUser()`, where these data are available. For now listening on this event and setting attributes in the token doesn't change anything, as it is the `PostAuthenticationGuardToken` that will always be stored.

A solution could be to add a `authenticatedToken` property to the authenticator, which could be set in `getUser()` and cleared in `createAuthenticatedToken` while returning its value (`JWTUserToken`,  potentially customized by the user), but it looks terribly wrong to me.

Any clue, please 😰 ?

Resources/doc/1-configuration-reference.md
chalasr: Not so big at the moment but [here](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/guard_jwt_authenticator/Resources/doc/6-extending-jwt-authenticator.md) you can read it :)

chalasr:
I plane to merge this one in the next days

PR #188, opened 2016-07-02
[DX] Add JWTNotFoundEvent

### LINKED ISSUE #187: lexik_jwt_authentication.on_jwt_invalid fired only for valid JWTS
Hello,
I want to log all invalid requests to my API so registered a listener: **lexik_jwt_authentication.on_jwt_invalid**

I have made some tests:  **lexik_jwt_authentication.on_jwt_invalid** is not fired when token does not meet JWT syntax requirements or is not defined in http header. Only **AuthenticationCredentialsNotFoundException** exception is logged in profiler:

`
Context: { "exception": "Object(Symfony\\Component\\Security\\Core\\Exception\\AuthenticationCredentialsNotFoundException)" } `

Is there any way to solve this in some easy way? Thank you for your time!

chalasr:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      |  no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #187, https://github.com/lexik/LexikJWTAuthenticationBundle/pull/159#issuecomment-208228006
| Tests pass?   | yes

As discussed in the fixed tickets,, I propose to add a specific `JWT_NOT_FOUND`, making users able to catch this event and set a custom response if they want, or just ignore it.
It can be useful to secure routes that are part of anonymous firewalls, for restrict it to a specific user role for instance, or just have the same behaviour as on `JWT_INVALID`.

Also it introduces an interface implemented by both `JWT_NOT_FOUND` and `JWT_INVALID`, so the same (listener) method can be used to listen on both events.

Closes #187.

### DIFF
--- a/Event/JWTFailureEventInterface.php
+++ b/Event/JWTFailureEventInterface.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Lexik\Bundle\JWTAuthenticationBundle\Event;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Core\Exception\AuthenticationException;
+
+/**
+ * Interface for event classes that are dispatched when a JWT cannot be authenticated.
+ * 
+ * @author Robin Chalas <robin.chalas@gmail.com>
+ */
+interface JWTFailureEventInterface
+{
+    /**
+     * @return Response
+     */
+    public function getResponse();
+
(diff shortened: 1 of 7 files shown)


chalasr:
@slashfan The issue fixed by this PR exists for a long time, plus it will helps for #184.
If you agree, after you reviewed it I think we can draft the `v1.6.0` release.

PR #195, opened 2016-07-09
Configuration options modified

Spomky:
This PR modifies the following configuration option:
* the `encryption_algorithm` is renamed into `signature_algorithm`.
Documentation is updated accordingly.

This PR includes minor modifications on the service.xml file.

PR #196, opened 2016-07-09
[Configuration] Make *_key_path config options not mandatory

chalasr:
Since we want to integrate different bridges of JW* libraries, the `*_key_path` options should not be mandatory anymore, because some of their libraries (especially https://github.com/spomky-labs/lexik-jose-bridge) doesn't need this to be configured.

@Spomky, you worked on this in #195 (partially), sorry for that, we will profit of this change to perform some checks (key files existence) via config validation. Let me know what do you think.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -24,10 +24,18 @@ class Configuration implements ConfigurationInterface
             ->addDefaultsIfNotSet()
             ->children()
                 ->scalarNode('private_key_path')
-                    ->cannotBeEmpty()
+                    ->defaultNull()
+                    ->validate()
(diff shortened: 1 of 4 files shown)


Spomky:
Looks good.
Just a question: is it relevant to check again the path in the KeyLoader class? It is already checked if the Configuration class.

chalasr:
Thank you for reviewing it.

> It is already checked if the Configuration class.

I hesitated to remove this additional check and I understand it doesn't feel relevant, but I thought it is preferable to keep it, in order to prevent the case of this value is overridden by the end user or some other edge cases of this kind. WDYT?

Spomky:
OK it makes sense.

PR #200, opened 2016-07-18
Depreciate injection of Request instances

chalasr:
| Q             | A
| ------------- | ---
| Branch?       | master
| BC breaks?    | yes (small)
| Deprecations? | yes
| Tests pass?   | yes

This depreciates the injection of `Request` instances in order to totally avoid this practice in 2.0. That concerns most of our Event classes.

Warning triggered by Scrutinizer:

> _Bug_ 
You have injected the Request via parameter $request. This is generally not recommended as there might be multiple instances during a request cycle (f.e. when using sub-requests). Instead, it is recommended to inject the RequestStack and retrieve the current request each time you need it via getCurrentRequest().

The proposed deprecation way consists into:

- adding a default value (null) for constructors/setters arguments type hinted as `Request` (BC break)
- document the corresponding properties+members as deprecated
- trigger a `E_USER_DEPRECATED` notice when the deprecated methods are called with a Request as argument.

The recommended alternative is to inject the RequestStack instead.
For `Event` classes, the user will be responsible of injecting the request stack into their event listeners to be able to use it.

All changed classes are instantiated internally, I think this should not be an hard BC breaking for most people.
Any thoughts? 

----------

- [x] Pre-remove the practice from our code (when RequestStack exists)
- [x] Documentation needs to be updated with an example of `@request_stack` injection&usage


chalasr:
Pending https://github.com/gesdinet/JWTRefreshTokenBundle/pull/45

PR #201, opened 2016-07-19
Remove deprecated request args

chalasr:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes

This totally remove the ability of injecting request instances in Event classes, plus there is no more need of injecting the `@request_stack` service into the `JWTManager` class.


Services/JWTManager.php
```diff
@@ -42,21 +41,14 @@ class JWTManager implements JWTManagerInterface
...
-     */
-    protected $requestStack;
-
-    /**
      * @param JWTEncoderInterface      $encoder
      * @param EventDispatcherInterface $dispatcher
      * @param int                      $ttl
-     * @param RequestStack             $requestStack
      */
-    public function __construct(JWTEncoderInterface $encoder, EventDispatcherInterface $dispatcher, RequestStack $requestStack, $ttl)
```
chalasr: One less BC break in 2.0.

PR #202, opened 2016-07-21
Exceptions simplified

Spomky:
This PR removes some exceptions as they do not add anything except a preset message.
Now only one class is left and should be used on every JWT encoding/decoding error.


chalasr:
Thank you for that @Spomky 👍 .

One thing though, I would have kept two kind of failures separated, i.e. `JWTEncodeFailureException` and  `JWTDecodeFailureException`, WDYT?

Spomky:
This means to have two classes with exactly the same content (except the name of the class).

chalasr:
Not a problem to me, exception classes are often empty, only the name change for a matter of easy catching and clear stack traces.

Spomky:
OK. I will update that PR

PR #224, opened 2016-08-21
Token TTL reduced

Spomky:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no|
| BC breaks? | no |
| Deprecations | no|
| Fixed tickets | no |
| Tests pass | ? |

86400 seconds (24 hrs) seem to be to high for a token TTL.
This commit recudes this value to 3600 seconds (1 hr).

1 hour should be enough for almost all use cases. If not, a bundle that
provides refresh token implementation will allow to issue another token
when it is expired.


Spomky:
Branch not rebased

chalasr:
👍  For me. 

However, this change is somehow a BC break, people who don't explicitly set this value in the past will get their tokens expire faster, so I would merge this in 2.0. @Spomky WDYT?

It's an important decision, I would prefer have the agreement of another (or more) maintainers.
ping @slashfan

PR #225, opened 2016-08-21
TTL scaled down

Spomky:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no|
| BC breaks? | yes |
| Deprecations | no|
| Fixed tickets | no |
| Tests pass | yes |

86400 seconds (24 hrs) seem to be too high for a token TTL.
This commit reduces this value to 3600 seconds (1 hr).

1 hour should be enough for almost all use cases. If not, a bundle that
provides refresh token implementation will allow to issue another token
when it is expired.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -41,7 +41,7 @@ class Configuration implements ConfigurationInterface
                     ->defaultValue('')
                 ->end()
                 ->scalarNode('token_ttl')
-                    ->defaultValue(86400)
+                    ->defaultValue(3600)
                 ->end()
                 ->arrayNode('encoder')
                     ->addDefaultsIfNotSet()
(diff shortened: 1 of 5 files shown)


Spomky:
This PR replace the #224. Sorry for the incovenient.
As mentionned in the previous PR, I noted there is a BC Break and should be merged (if accepted) with the branch 2.0.

chalasr:
Thank you for this @Spomky. 
Note that you can [now switch the base of any PR](https://github.com/blog/2224-change-the-base-branch-of-a-pull-request) :)

> 👍  For me. 

> However, this change is somehow a BC break, people who don't explicitly set this value in the past will get their tokens expire faster, so I would merge this in 2.0. @Spomky WDYT?

> It's an important decision, I would prefer have the agreement of another (or more) maintainers.
ping @slashfan

chalasr:
I'm personally in favor of giving default values as close as possible to the minimum, especially when we talk about security and, the main maintainer of this bundle being quite busy these times (@slashfan), let's merge this for now.

Thank you Florent.

PR #226, opened 2016-08-22
Enhance naming

chalasr:
Next to #195 where we changed `encryption_algorithm` to `signature_algorithm`, this adapt the (forgotten) existing code, changing `encryptionAlgorithm` to `signatureAlgorithm` everywhere. 

2.0 being not yet released, we must profit to ensure that the introduced features are following good practices, conventions and naming.

@Spomky as you are quite aware of the good practices in term of encoding in general, it would be awesome if you could do a quick review of the terms we use in the corresponding part of our code base, especially on the `2.0` branch after #162.

For instance I thought about `encryption_engine`, is it always correct in the contexts we use it?
Otherwise, what about simply name it `engine`? That would give `encoder.engine` as option name.

Let me know about your suggestions!

### DIFF
--- a/Services/JWSProvider.php
+++ b/Services/JWSProvider.php
@@ -30,28 +30,28 @@ class JWSProvider implements JWSProviderInterface
     /**
      * @var string
      */
-    private $encryptionAlgorithm;
+    private $signatureAlgorithm;
 
     /**
      * @param KeyLoaderInterface $keyLoader
      * @param string             $encryptionEngine
-     * @param string             $encryptionAlgorithm
+     * @param string             $signatureAlgorithm
      *
      * @throws \InvalidArgumentException If the given algorithm is not supported
      */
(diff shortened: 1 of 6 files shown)


Spomky:
Hi,

Even if I am not sure to be fully qualified for that, let try to review that wording.

From my POV, PHPSecLib and OpenSSL are cryptographic libraries, so `encryption_engine` is correct however `crypto_engine` should be more appropriate.

I noted some PHPDoc information to be updated accordingly or such as in [KeyLoaderInterface.php#L6](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/2.0/Services/KeyLoader/KeyLoaderInterface.php#L6): Interface for classes that are able to load ~~SSL encryption~~ keys.

Anyway, method names are quite clear and should not be changed.

chalasr:
> Interface for classes that are able to load ~~SSL encryption~~ keys.

Would you have a suggestion of correction?  `"... load crypto keys"`?

Thank you @Spomky 👍

PR #227, opened 2016-08-22
Allow choosing a JWT encoding algo

kix:
Perhaps the next step would be adding support for different PHP JWT implementations.


chalasr:
Hi @kix,

Sorry for the confusion but we are currently working on a `2.0` version (see #132 for all the achievements) where this feature has already been implemented. You will now be able to choose between all the algorithms you implemented here, plus you can choose to use phpseclib insteadof openssl, and many other cool stuff that have been implemented (or are in progress, see #132,  the opened pulls request and [the documentation on 2.0](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/2.0/Resources/doc/5-encoder-service.md) which is updated in sync with the branch).

I'll add a README note on the master branch about that `2.0` is in dev (and maybe master feature frozen).

Thank you for your effort.

kix:
@chalasr, thanks for the notice! Well, how do you like the idea of allowing the end user to pick a specific decoder implementation somehow? That was the idea I was planning to use for my another PR here :)

kix:
Ah, right, I see, you have that feature there already.

chalasr:
:) Hope you like it! Thank you again @kix

chalasr:
Note that you can use your own encoder based on the library of your choice on the current stable `v1.7`

PR #237, opened 2016-09-20
Introduce a FormLoginAuthenticator

chalasr:
| Q             | A    |
|---------------|------|
| Bug fix?      | no  |
| New feature?  | yes |
| BC breaks?    | not for now |
| Deprecations | yes |
| Fixed tickets | helps in #132 |
| Tests pass?   | yes  |

For now I kept the authentication success/failure listeners as is, injected into the authenticator. To me it's fine to don't move their logic in the authenticator itself, so the classic `form_login` (or whatever) can just continue to work. Plus, `GuardAuthenticatorInterface` doesn't extend the corresponding interfaces so I guess it's not a problem. Please stop me if I'm wrong.

-----

- [x] Add the authenticator
- [x] Update functional tests for using it
- [ ] Add a unit test
- [ ] Update the doc

### DIFF
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -34,6 +34,15 @@
         <!-- Default JWT Token Authenticator -->
         <service id="lexik_jwt_authentication.jwt_token_authenticator" parent="lexik_jwt_authentication.security.guard.jwt_token_authenticator" />
 
+        <!-- Abstract FormLogin Authenticator / Guard implementation -->
+        <service id="lexik_jwt_authentication.security.guard.form_login_authenticator" class="Lexik\Bundle\JWTAuthenticationBundle\Security\Guard\FormLoginAuthenticator" abstract="true">
+            <argument type="service" id="security.password_encoder" />
+            <argument type="service" id="lexik_jwt_authentication.handler.authentication_success" />
+            <argument type="service" id="lexik_jwt_authentication.handler.authentication_failure" />
+        </service>
+        <!-- Default FormLogin Authenticator -->
+        <service id="lexik_jwt_authentication.form_login_authenticator" parent="lexik_jwt_authentication.security.guard.form_login_authenticator" />
+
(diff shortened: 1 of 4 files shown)


chalasr:
Still need to determine if the `check_path`, `username_parameter` and `password_parameter` need to be set from the main configuration, or if we just use the values hard set in the authenticator and let the developer override it directly (too much work IMHO as only one authenticator will be used in most cases).

chalasr:
I would like to share my pov on this, in order to be sure it is worth continuing:

- The built-in `form_login` doesn't involve any logic from our side, except the two authentication success/failure listeners that would be kept anyway for a matter of flexibility.
- It has a factory with a structured configuration tree, so it doesn't involve to configure anything  from the bundle itself
- Some nice stuff can be easily implemented by extending/replacing it (e.g. a json implementation as the one which may be integrated to symfony in https://github.com/symfony/symfony/pull/18952 or a stateless replacement as the one provided by https://github.com/gfreeau/GfreeauGetJWTBundle, and we could eventually create our own factory on top of the `form_login` one instead of this)

On the other hand:

- This needs to add some `security` specific configuration in the bundle configuration including `username_parameter`, `password_parameter` and `check_path`, that is weird IMHO
- This just does what the `form_login` already does, without any gain in term of line of code/config for the end user (worst: this may involve more for the same result)

Contrarily to this, the `JWTTokenAuthenticator` solves a very specific need and removes some complexity from our code, IMHO it is far more useful to have as a guard authenticator. 
So, here is my thought: If an user wants to use Guard for getting its token, then he just have to create the authenticator himself, _rather than_ writing configuration (hard-defined values are good enough for an application, not for third party tools)

@slashfan I need your thought on this for progressing on the v2.
@weaverryan It would be nice if we can have your opinion here.

Any opinion is welcome.

chalasr:
Closing due to the lack of comments.

PR #238, opened 2016-09-22
Fix iat claim

chalasr:
Since #162 our encoder uses `Namshi\Jose\JWS` instead of `Namshi\Jose\SimpleJWS`. I just spotted that only `SimpleJWS` set the `iat` claim.

However, I am not sure any check is done on this claim, so maybe it should be done from our encoder, as a bug fix on master (1.x). 
ping @Spomky do you know if I'm wrong?

### DIFF
--- a/Encoder/DefaultEncoder.php
+++ b/Encoder/DefaultEncoder.php
@@ -32,6 +32,8 @@ class DefaultEncoder implements JWTEncoderInterface
      */
     public function encode(array $payload)
     {
+        $payload['iat'] = time();
+
         try {
             $jws = $this->jwsProvider->create($payload);
         } catch (InvalidArgumentException $e) {


Spomky:
There is no check on that claim. AFAIK, same goes for the other claims including the `exp` one.
The [jwt.io](https://jwt.io/) page indicates that checks are performed on `exp`, `nbf` and `iat` but from my POV it is not true.
You have to check those claims (that is what you do with the [`LoadedJWS`](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/2.0/Signature/LoadedJWS.php#L70) class).

In that PR you add the `iat` claim, however you have to verify it somewhere. The `LoadedJWS` class should manage that.

chalasr:
Yeah it was my guess after a quick look... thank you for confirm me

PR #240, opened 2016-09-23
Add iat check

### LINKED ISSUE #239: Add iat check
`namshi/jose` adds the `iat` claim to the token payload but it doesn't check it when decoding the token, we should do it from our side (as ~~bugfix on master~~ feature on 2.0 for BC).

chalasr:
| Q             | A    |
|---------------|------|
| Bug fix?      | no  |
| New feature?  | yes |
| BC breaks?    | no |
| Deprecations | no |
| Fixed tickets | #239 |
| Tests pass?   | yes  |

Closes #239


chalasr:
We'll keep the claim not checked on 1.x, it would be somehow a BC break, e.g. for people who don't use our built-in encoder and build their own payloads without this claim.

PR #249, opened 2016-10-15
Avoid setting exp claim from JWTManager

chalasr:
| Q             | A    |
|---------------|------|
| Bug fix?      | no  |
| New feature?  | yes |
| BC breaks?    | yes |
| Deprecations | no |
| Fixed tickets | n/a |
| Tests pass?   | yes  |

The encoder should be responsible of setting the `exp` claim and all other standard ones. @Spomky this might have an impact on the `lexik_jose_bridge`, excepted if you set this claim from the encoder anyway.

- [x] Inject the configured `ttl` in `*JWSProvider` rather than `JWTManager`
- [x] Update tests
- [ ] Add an UPGRADE+CHANGELOG note

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,9 @@ For a diff between two versions https://github.com/lexik/LexikJWTAuthenticationB
 
 ## [2.0](https://github.com/lexik/LexikJWTAuthenticationBundle/tree/2.0)
 
-* feature [\#246](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/247) Add a simple built-in encoder based on lcobucci/jwt ([chalasr](https://github.com/chalasr))
+* feature [\#249](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/249) Avoid setting exp claim from JWTManager ([chalasr](https://github.com/chalasr))
+
+* feature [\#246](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/246) Add a simple built-in encoder based on lcobucci/jwt ([chalasr](https://github.com/chalasr))
 
 * feature [\#240](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/240) Add iat check ([chalasr](https://github.com/chalasr))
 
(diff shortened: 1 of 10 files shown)


Spomky:
The lexik-jose-bridge bundle [marks the `exp` claim as critical](https://github.com/Spomky-Labs/lexik-jose-bridge/blob/master/Encoder/LexikJoseEncoder.php#L214).

It does not include it as it seems that it is already set in the payload [by the JWTManager service](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/2.0/Services/JWTManager.php#L64).

chalasr:
Is the change itself fine to you?
It sounds logical to me from a separation of concerns pov and allows to use the encoder directly without bad surprise.

Spomky:
Yes it is. The payload sent by the JWTManager should contain only the claims related to the user authentication.
Other claims such as `jti`, `iat`, `exp` and so on are not related to the user but the JWT creation/loading.

DependencyInjection/Configuration.php
```diff
@@ -42,6 +42,12 @@ public function getConfigTreeBuilder()
                 ->end()
                 ->scalarNode('token_ttl')
                     ->defaultValue(3600)
+                    ->validate()
+                        ->ifTrue(function ($ttl) {
+                            return !is_numeric($ttl);
```
lsmith77: we should allow `null` values here as per https://github.com/lexik/LexikJWTAuthenticationBundle/pull/117

lsmith77: also why not simply make it an `integerNode` as I suspect that floats do not really make sense here ..

chalasr: this check has been moved at runtime in order to allow use of `%env()%` parameters in https://github.com/lexik/LexikJWTAuthenticationBundle/pull/325. The point of using `is_numeric()` is to support env vars too.
About allowing null, this has been changed accidentally and rediscussed in https://github.com/lexik/LexikJWTAuthenticationBundle/issues/250, having tokens with an infinite lifetime is not something we want to support regarding security

CHANGELOG.md
```diff
@@ -15,7 +17,7 @@ For a diff between two versions https://github.com/lexik/LexikJWTAuthenticationB
 
 * feature [\#218](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/218) Add more flexibility in token extractors configuration ([chalasr](https://github.com/chalasr))
 
-* feature [\#217](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/217) Refactor TokenExtractors loading for easy overriding ([chalasr](https://github.com/chalasr))
+* feature [\#217](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/217) Refactor TokenExtractors loadi ng for easy overriding ([chalasr](https://github.com/chalasr))
```
guilliamxavier, line 20: Reverted in #750

Issue #256, opened 2016-10-20
Mark the exik_jwt_authentication.encoder.default service as autowire default for JWTEncoderInterface


weaverryan:
First, congrats on 2.0! I'm looking forward to checkout out all the changes!

On KnpU, since we use this bundle, I just had one user mention a small change in 2.0 that I think we can make smoother. Specifically, we use autowiring on a service that has `JWTEncoderInterface` as an argument. With 2.0, this gives the following exception:

>
Unable to autowire argument of type "Lexik\Bundle\JWTAuthenticationBundle\Encoder\JWTEncoderInterface" for the service "app.security.jwt_token_authenticator". Multiple services exist for this interface (lexik_jwt_authentication.encoder.default, lexik_jwt_authentication.encoder.lcobucci).

Could we mark the `lexik_jwt_authentication.encoder.default` as the correct autowiring service for this interface? We would just need an `autowiring_types` key added to that service (http://symfony.com/doc/current/components/dependency_injection/autowiring.html) - this is done in other places, like MonologBundle to mark `@logger` as the default `LoggerInterface` service.

Thanks!

chalasr:
@weaverryan Really, thank you very much for the help you given on this major version.
I'm glad to see you upgrading and I hope it will not be too much work. 

About the autowiring types, it's indeed something good to have that I totally missed 👍 
As you spotted the change has been made. I just tagged a `v2.0.1` so you should be able to get it.

I have to remind you that this bundle now has a built-in Guard authenticator, greatly inspired from the one of your KnpUniversity lesson and presentation (I'm pretty sure it is the very same that the DIC was not able to autowire :))

Feedbacks from you and your collaborators will be greatly appreciated.

Closing as fixed in #257. Thank you!

weaverryan:
Wow, that was so fast! And I already replied to the user (who just posted the comment this morning!). You rock :).

I have an issue internally to pull down 2.0.1 with our tutorial and make sure things are smooth - that'll be my chance to really test out the new version. I'm also excited to look at the authenticator - I will probably at least add a note so that users know that this is an option (versus building their own, which is a nice exercise, but not necessarily the best actual solution).

Cheers!

PR #257, opened 2016-10-20
Set autowiring types on services with many alternatives like encoder

chalasr:
See #256

### DIFF
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -10,6 +10,7 @@
         </service>
         <service id="lexik_jwt_authentication.encoder.default" parent="lexik_jwt_authentication.encoder.abstract">
             <argument type="service" id="lexik_jwt_authentication.jws_provider.default"/>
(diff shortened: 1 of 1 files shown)


weaverryan:
👍 this makes sense (and thanks, that was crazy fast)! The other 2 you added, for `JWSProviderInterface` and `TokenExtractorInterface` also seem to make sense, but it just depends on whether or not you want to encourage people to inject these directly into their services (I noticed that they're public=false). That's a subjective call :).

Cheers!

chalasr:
> it just depends on whether or not you want to encourage people to inject these directly into their service

For the `TokenExtractorInterface`, it needs to be injected into the `JWTTokenAuthenticator` which is intended to be extended by users. 

About the `JWSProviderInterface`, we realized that some needs (more or less specific) may involve to extend our `DefaultEncoder` with, sometimes, no need for re-writing the logic of the `DefaultJWSProvider` which is injected into, that's why I thought adding the autowired type onto may be useful.

Thanks for your review!

PR #260, opened 2016-10-25
Remove abstract definition for Guard JWTokenAuthenticator

santakadev:
Guard JWTokenAuthenticator class is not abstract but in service.xml is defined as abstract.

### DIFF
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -40,7 +40,7 @@
             <argument>%lexik_jwt_authentication.token_ttl%</argument>
         </service>
         <!-- Abstract JWT Token Authenticator / Guard implementation -->
-        <service id="lexik_jwt_authentication.security.guard.jwt_token_authenticator" class="Lexik\Bundle\JWTAuthenticationBundle\Security\Guard\JWTTokenAuthenticator" abstract="true">
+        <service id="lexik_jwt_authentication.security.guard.jwt_token_authenticator" class="Lexik\Bundle\JWTAuthenticationBundle\Security\Guard\JWTTokenAuthenticator">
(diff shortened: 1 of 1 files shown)


chalasr:
Hi @santakadev,

I'm not sure to understand the motivation of this PR.
Making a service definition abstract doesn't involve to have the corresponding class abstract, it is two different concepts. 
The purpose of making this definition abstract is to allow extending the service without having to repeat all of the arguments. Plus, this service will be removed from the container at compile time so nobody can neither inject it nor get it.
See https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Resources/doc/6-extending-jwt-authenticator.md#creating-your-own-token-authenticator for an example and http://symfony.com/doc/current/service_container/parent_services.html

However, we are setting the `class` on this service, making the `abstract` attribute quite useless. If we would make a change, it would be to move the `class` from the abstract `lexik_jwt_authentication.guard.jwt_token_authenticator` to the concrete `lexik_jwt_authentication.jwt_token_authenticator`. Would you consider changing the scope of this PR for doing that?

Thank you.

santakadev:
Hi @chalasr,

The motivation for this PR is that I was trying to use the first service definition of the guard authenticator in **security.yml** `lexik_jwt_authentication.guard.jwt_token_authenticator`.

I don't realize that there was another service definition extending the first. Using `lexik_jwt_authentication.jwt_token_authenticator` all works as expected.

Anyway, I don't understand why to make the service abstract, and then create another definition for the same class. We can extend the first definition without repeat the arguments, and an instance of the class is keeping in the container, because the second definition. Probably I am missing something.

Thanks for all.

PR #261, opened 2016-10-25
The security token must be authenticated no matter of the user's roles

### LINKED ISSUE #259: JWT Token not found returned when user doesn't have roles
With users that have none user roles attached I get 'JWT Token not found' when I'm making any API requests. It seems that 

/jwt-authentication-bundle/Security/Guard/JWTTokenAuthenticator.php
```
$authToken = new JWTUserToken($user->getRoles(), $user, $preAuthToken->getCredentials(), $providerKey);
```

/jwt-authentication-bundle/Security/Authentication/Token/JWTUserToken.php
```
/**
 * {@inheritdoc}
 */
public function __construct(array $roles = [], UserInterface $user = null, $rawToken = null, $providerKey = null)
{
    parent::__construct($roles);

    if ($user) {
        $this->setUser($user);
    }

    $this->setRawToken($rawToken);
    $this->setAuthenticated(count($roles) > 0);

    $this->providerKey = $providerKey;
}
```

Is causing the problem. Why are roles required on this case ?

chalasr:
Closes #259

### DIFF
--- a/Security/Authentication/Token/JWTUserToken.php
+++ b/Security/Authentication/Token/JWTUserToken.php
@@ -35,7 +35,7 @@ class JWTUserToken extends AbstractToken implements GuardTokenInterface
         }
 
         $this->setRawToken($rawToken);
-        $this->setAuthenticated(count($roles) > 0);
+        $this->setAuthenticated(true);
 
         $this->providerKey = $providerKey;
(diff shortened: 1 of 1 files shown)


tarlepp:
When is this released?

PR #276, opened 2016-11-29
Adding `getProviderKey()` to JWTUserToken

eXtreme:
In some situations it is useful to get the security provider key from the token. In fact, JWTUserToken is constructed with `$providerKey` but it is never used and there is no getter for it, hence my addition.
It is also consistent with Symfony's `UsernamePasswordToken`, `RememberMeToken` or `PreAuthenticatedToken` which all have such getter.


chalasr:
Good catch! Thank you @eXtreme

PR #279, opened 2016-12-02
Refactor travis build for testing on lowest/highest sf version

chalasr:
- [ ] Update the README

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,25 +4,17 @@ php:
   - 5.5
   - 5.6
   - 7.0
+  - 5.5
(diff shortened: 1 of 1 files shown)


Spomky:
I use the same configuration on my projects.
See https://github.com/Spomky-Labs/jose-bundle/blob/master/.travis.yml for example.

chalasr:
Finally I think it's legit to keep at least one build per main symfony version on our lowest supported PHP version (5.5), and build on lowest/highest for other. See #280, let me know wdyt :)

PR #280, opened 2016-12-02
Build on symfony 3.2 in travis

chalasr:


chalasr:
Fixed the build on hhvm-stable btw.

PR #285, opened 2016-12-04
Avoid validating key paths before container compilation

### LINKED ISSUE #284: Path validation does not work with StreamWrappers
Since version 2 there is a check n the configuration if the paths for private and public key are valid. This is a good thing, because you can show an error message to the user already while the container is being compiled. However, it doesn't work if you use stream wrappers. I use a stream wrapper (AWS S3) in my production environment, so that all nodes can access the same key pair. These wrappers are only initialized after the container is compiled. Because of that the check in the configuration fails and I cannot clean and warmup the cache during composer install in my production environment.
Is there a way to fix this?

Best,
Chris

chalasr:
Fixes #284 by removing key path validation from our configuration since it's too early for stream wrappers. Paths are still checked at runtime on credentials+jwt login.

/cc @Spomky maybe `SpomkyLabs/JoseBundle` is concerned by this issue too

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -25,17 +25,9 @@ class Configuration implements ConfigurationInterface
             ->children()
                 ->scalarNode('private_key_path')
                     ->defaultNull()
-                    ->validate()
-                    ->ifTrue($this->isPathInvalid())
-                        ->thenInvalid(self::INVALID_KEY_PATH)
-                    ->end()
                 ->end()
                 ->scalarNode('public_key_path')
                     ->defaultNull()
-                    ->validate()
-                    ->ifTrue($this->isPathInvalid())
-                        ->thenInvalid(self::INVALID_KEY_PATH)
-                    ->end()
                 ->end()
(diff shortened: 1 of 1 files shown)


Spomky:
Hi @chalasr,

I am not sure the [spomky-labs/jose-bundle](https://github.com/Spomky-Labs/jose-bundle) is concerned as [there is no path validation in the key configuration](https://github.com/Spomky-Labs/jose-bundle/blob/master/DependencyInjection/Source/JWKSource/KeyFile.php#L57). The bundle only creates the key definitions using [a dedicated method](https://github.com/Spomky-Labs/jose-bundle/blob/master/DependencyInjection/Source/JWKSource/KeyFile.php#L25) called by the compiler pass. The path is checked only when the definition is used and the key factory is called.

However, I will be happy to know if I am wrong or not a fix it if needed.
Please @cevou, could you try to use the [spomky-labs/lexik-jose-bridge](https://github.com/Spomky-Labs/lexik-jose-bridge) and verify this assumption?

PR #287, opened 2016-12-15
Response failure content, avoid override existing properties

kevin-lot:


chalasr:
Good catch! thank you @kevin-lot

chalasr:
V2.1.0 has been released and include this change

PR #290, opened 2016-12-19
Delete duplicate link into "Further documentation"

anivon:
Delete duplicate "Extending JWTTokenAuthenticator" link into "Further documentation" section


chalasr:
Nice catch, thank you @anivon

PR #296, opened 2016-12-29
Add quotes to parameters in YML

ruudk:
For the copy-pasters :)


chalasr:
Nice catch! Thank you @ruudk

PR #297, opened 2016-12-30
Remove $ prefix for easy copy paste

ruudk:


chalasr:
Thank you for your contribution @ruudk, it's much appreciated. 

However, I think I don't agree with this change for the following reasons:
- Having the `$` eases readability: it's clear that it's a CLI command to run
- Yet it's a bit annoying to remove the `$` after for copy-pasting, but I do think it is not that bad, because running commands should be done carefully. Plus the commands updated here must be ran separately as one of them needs some interactions, so pasting the whole block in the CLI would be annoying as well

ruudk:
I see your point. Too bad :)

PR #302, opened 2017-01-20
JWTUserProvider refreshUser method must return a User object

MisterGlass:
According to the [UserProviderInterface definition](http://api.symfony.com/3.2/Symfony/Component/Security/Core/User/UserProviderInterface.html#method_refreshUser), the refreshUser method should return an object implementing UserInterface.

This was returning null & causing an error when trying to chain multiple providers together.

### DIFF
--- a/Security/User/JWTUserProvider.php
+++ b/Security/User/JWTUserProvider.php
@@ -51,6 +51,6 @@ final class JWTUserProvider implements UserProviderInterface
 
     public function refreshUser(UserInterface $user)
     {
-        // noop
+        return $user; // noop
     }
 }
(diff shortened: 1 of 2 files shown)


chalasr:
Good catch! Thank you @MisterGlass

MisterGlass:
That was quick! Any chance you can tag a release on this? I had to set my composer to dev stability to grab it.

chalasr:
@MisterGlass Done! You can upgrade to [v2.1.1](https://github.com/lexik/LexikJWTAuthenticationBundle/releases/tag/v2.1.1)

MisterGlass:
Thanks!

PR #309, opened 2017-02-27
Fixed a typo in AuthenticationFailureEvent

blaues0cke:


chalasr:
Good catch. Thank you @blaues0cke

PR #311, opened 2017-02-28
Handle empty or null authorization header prefix, closes #310

### LINKED ISSUE #310: Handle empty or null header prefix
First, thank you for your great bundle!

`AuthorizationHeaderTokenExtractor->extract` method works well for non-null/empty prefixes (`Bearer` for example), but will return `false` otherwise.

For instance, I'd like to use a `X-Auth-Token: {token}` header. I set this up in my `config.yml`:
```yaml
token_extractors:
    authorization_header:
        prefix: ~
        name: X-Auth-Token
```

In this case, the [explode](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/TokenExtractor/AuthorizationHeaderTokenExtractor.php#L43) function will return an array with a single element (the token): we should check if the `prefix` is either null or empty before blindly assuming it's not, shouldn't we?

chteuchteu:
This PR handles cases where `token_extractors.authorization_header.prefix` is null (`~`) or empty (`""`)

```yaml
token_extractors:
    authorization_header:
        prefix: ~
        name: X-Auth-Token
```

### DIFF
--- a/TokenExtractor/AuthorizationHeaderTokenExtractor.php
+++ b/TokenExtractor/AuthorizationHeaderTokenExtractor.php
@@ -22,8 +22,8 @@ class AuthorizationHeaderTokenExtractor implements TokenExtractorInterface
     protected $name;
 
     /**
-     * @param string $prefix
-     * @param string $name
+     * @param string|null $prefix
+     * @param string      $name
      */
     public function __construct($prefix, $name)
     {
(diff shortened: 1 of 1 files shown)


chteuchteu:
@chalasr : should we use a strict comparison (`$this->prefix === null || $this->prefix === ''`) instead?

chalasr:
Could you please take care of the StyleCI failure? Also rebase this on our 1.x branch?

chalasr:
Thank you @chteuchteu

PR #312, opened 2017-03-02
Ease sharing keys between parties

### LINKED ISSUE #305: Access public key string programmaticaly
In order to build extensible communications I want to publish my public key from signing key-pair as a separate controller file.

If I try to use the following code

```php
    public function pubkeyAction()
    {
        $key = $this->get('lexik_jwt_authentication.key_loader')->loadKey('public');
        dump($key);

        return new Response(
            stream_get_contents($key),
            200,
            ['Content-Type' => 'text/plain']
        );
    }
```

it shows me 

```
Warning: stream_get_contents(): supplied resource is not a valid stream resource
```

Is there any convenient way to export public key without knowing the nature of underlying library that reads the key from the filesystem?

I have an option to manually fetch the key from the container parameters, but I think this is not a good choice

chalasr:
Closes #305.

@Spomky A review from you would be really appreciated.

/cc @scaytrase

### DIFF
--- a/Services/KeyLoader/KeyDumperInterface.php
+++ b/Services/KeyLoader/KeyDumperInterface.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Lexik\Bundle\JWTAuthenticationBundle\Services\KeyLoader;
+
+/**
+ * @author Robin Chalas <robin.chalas@gmail.com>
+ */
+interface KeyDumperInterface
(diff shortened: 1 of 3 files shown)


Spomky:
Hi @chalasr 

This PR looks good to me.

chalasr:
Thank you @Spomky!

PR #314, opened 2017-03-23
Update 3-functional-testing.md

MasseElch:
Typo in ssl-key-paths

### DIFF
--- a/Resources/doc/3-functional-testing.md
+++ b/Resources/doc/3-functional-testing.md
@@ -16,8 +16,8 @@ Override the bundle configuration in your `config_test.yml` :
 ``` yaml
 # config_test.yml
 lexik_jwt_authentication:
-    private_key_path:   %kernel.root_dir%/../var/jwt/private-test.pem
-    public_key_path:    %kernel.root_dir%/../var/jwt/public-test.pem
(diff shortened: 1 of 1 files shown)


chalasr:
Good catch, thank you @MasseElch

chalasr:
I just realize that this is wrong, the doc was correct. Keys should go in the `var/jwt` directory, not `app/var/jwt`. Reverting.

chalasr:
See https://github.com/lexik/LexikJWTAuthenticationBundle/pull/319

MasseElch:
The `var` directory will not go into git, but for our CI-Pipeline we created some keys only for testing purposes.

I thought this was your intention when you wrote the following lines for generating a testing key-pair:

```
$ openssl genrsa -out app/var/jwt/private-test.pem -aes256 4096
$ openssl rsa -pubout -in app/var/jwt/private-test.pem -out app/var/jwt/public-test.pem
```

MasseElch:
Ah nevermind, you just did.

chalasr:
@MasseElch Yeah, that's the mistake. `app` was the right place before symfony 3. Sorry for the noise and thanks again for pointing the error.

PR #316, opened 2017-03-29
Share Token Extractor Config via parameters

Batch1211:
I stumbled upon this issue, when I tried to make the LexikJWTAuthenticationBundle work with cookies. But I'll try to explain the issue in general terms:

I think it is a good idea to share the config parameters for the token extractors with other services and bundles via the DI container. It might be very useful for other bundles to know, under what name the CookieTokenExtractor is looking for the JWT string in the request. The same goes for the Authorization header and the Query String extractor.

So I propose to share these config namespaces in the DI container, the same way the friends of symfony do. They have created these two nifty little helper methods (`remapParameters()` & `remapParametersNamespaces()`) that let us easily write config values in the parameter namespace.

### DIFF
--- a/DependencyInjection/LexikJWTAuthenticationExtension.php
+++ b/DependencyInjection/LexikJWTAuthenticationExtension.php
@@ -58,6 +58,9 @@ class LexikJWTAuthenticationExtension extends Extension
                 ->replaceArgument(1, $tokenExtractorsConfig['authorization_header']['name']);
 
             $map[] = new Reference($authorizationHeaderExtractorId);
+            $this->remapParametersNamespaces($tokenExtractorsConfig, $container, [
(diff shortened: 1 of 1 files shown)


chalasr:
@Batch1211 First of all, thank you for contributing.
I'm against the proposed change first because adding parameters involve to maintain them: There is no mechanism to deprecate parameters. If we don't need them a day, we'll have to break backward compatibility for people relying on these parameters. 
About the global idea, I don't see much value in having application related configuration available in the container. The DI container is dumped wh

chalasr:
Sorry for closing, it's a mistake.

Batch1211:
@chalasr Thanks for your reply. I see your points and didn't know about the PrependExtensionInterface. I will use this instead.

chalasr:
@Batch1211 Thanks for your understanding.

PR #319, opened 2017-03-30
Fix key path in functional testing doc

chalasr:
Superseds #314


carlcraig:
I think the paths are incorrect, should most likely be `var` instead of `var/var`

chalasr:
@carlcraig fixed in https://github.com/lexik/LexikJWTAuthenticationBundle/commit/5960e82097c4d87f1621b293cb526b0b2b6bd50a, thank you.

Issue #313, opened 2017-03-08
Moving token from body to header


IljaN:
I would like to have the token which is returned after login, to be returned in the header instead of the body.

I created a AuthenticationSuccessListener, but can`t overide the body to make it empty. Is this intentional?
```php
    public function onAuthenticationSuccessResponse(AuthenticationSuccessEvent $event) {
        $data = $event->getData();

        if (!$event->getUser() instanceof UserInterface) {
            return;
        }

        $response = $event->getResponse();
        $response->headers->add(['authorization' => $data['token']]); // works
        $response->setContent(''); // no effect
        $event->setData([]); // no effect
    }
```

chalasr:
This is the expected behavior indeed, but that's because we didnot see any valid use case at the moment it has been implemented. I'll have a look at what we can do, maybe this event deserves a `setResponse()` as well.

IljaN:
My usecase is that for example vue-auth defaults to reading the token from the header after login. I googled and found some other jwt implementations that also behave like that altough they seem to be in the minority.

chalasr:
Fixed by #320

IljaN:
@Batch1211 Thank you very much! <3

PR #320, opened 2017-03-30
feat: Allow for Response Body without JWT Token

Batch1211:
This pull request is regarding Issue #313.

My proposed change would not change to current behavior of the bundle. All it does is respecting any changes being made to the response data during the AUTHENTICATION_SUCCESS event. That is, if an event listener decides to remove the token from the data array, it will no longer be added later.

### DIFF
--- a/Response/JWTAuthenticationSuccessResponse.php
+++ b/Response/JWTAuthenticationSuccessResponse.php
@@ -26,20 +26,8 @@ final class JWTAuthenticationSuccessResponse extends JsonResponse
      * @param string $token Json Web Token
      * @param array  $data  Extra data passed to the response
      */
-    public function __construct($token, array $data = null)
+    public function __construct($token, array $data = [])
     {
-        $this->token = $token;
-
-        parent::__construct($data);
-    }
-
-    /**
-     * Sets the response data with the JWT included.
-     *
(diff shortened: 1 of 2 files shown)


chalasr:
👍 
Could you have a look at the failing test? It surely ensures that the token can't be removed from the response data, that is no more relevant.

Batch1211:
Looking at the tests, I realized, that a better implementation would be to add the token to the response data in the constructor. This brings the implementation closer to the original JsonResponse implementation while preserving the previous behavior.

The replaceData Test now checks that the token is really removed, when the data array does not contain the token data.

Response/JWTAuthenticationSuccessResponse.php
```diff
@@ -28,18 +28,8 @@
      */
     public function __construct($token, array $data = null)
     {
-        $this->token = $token;
+        $data = is_null($data) ? [] : $data;
```
chalasr: Could `$response = null` be changed to `$response = []` in the signature? It would remove the need for this line

Batch1211: Done.

chalasr:
Thank you @Batch1211!

PR #325, opened 2017-04-14
Fix: Move `is_numeric` check from build time to runtime

### LINKED ISSUE #323: "token_ttl" must be a numeric value, but that's not compatible with Sf 3.2+ `%env(FOO)%` mechanism
Hi,

As we now have this shiny `%env(FOO)%` mechanism in Symfony configuration [for a few months](http://symfony.com/blog/new-in-symfony-3-2-runtime-environment-variables), and as it looks like it will be omnipresent in Symfony 4, it would be nice to be able to use it with this (wonderful) Bundle, wouldn't it?

But if I'm not wrong, it's not possible at the moment, because of [this check](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/DependencyInjection/Configuration.php#L39) in the Bundle's semantic configuration.

From what I understood from [that Symfony PR](https://github.com/symfony/symfony/pull/19681), these `%env(FOO)%` placeholders are actually replaced with placeholders (which are strings) at container compilation time, and will be replaced with the matching environment variables values at runtime.

Because of this, with the following config:
```yaml
lexik_jwt_authentication:
  token_ttl: '%env(JWT_TOKEN_TTL)%'
```

...I get the following at container build time:
ttl-is-not-numeric-at-compilation-time

Hence the validation fail.

Maybe it would be worth removing that `is_numeric()` check, so that we could actually configure our tokens TTL via environment variables?
If you're ok with that, I should be able to submit a PR soon.

Thanks!

DrBenton:
This will allow us to use the Symfony 3.2+ "%env(JWT_TOKEN_TTL)%" syntax

see #323 for detailed description and discussion
Closes #323


chalasr:
Great! Thank you @DrBenton

PR #337, opened 2017-05-29
firewall provider name fixed.

ugurerkan:
Hello,

I noticed little errors on documentation when trying to implement JWT my app and just updated for others don't struggle to figure out.

- Provider name was wrong, updated based on[ 1-configuration-reference.md#security-configuration](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Resources/doc/1-configuration-reference.md#security-configuration). 
- Syntax error fixed on User class example.


chalasr:
Just removed a copy-pasted comment that is irrelevant in this section.
Thanks @ugurerkan!

PR #344, opened 2017-06-09
Add an extension point on the UserProviderWithPayloadSupportInterface

sroze:
So anybody can plug their own user provider and get access to the payload :)


chalasr:
Great! One thing: PHP5 is still supported for this major :)

Security/Guard/JWTTokenAuthenticator.php
```diff
@@ -17,6 +17,7 @@
 use Lexik\Bundle\JWTAuthenticationBundle\Security\Authentication\Token\JWTUserToken;
 use Lexik\Bundle\JWTAuthenticationBundle\Security\Authentication\Token\PreAuthenticationJWTUserToken;
 use Lexik\Bundle\JWTAuthenticationBundle\Security\User\JWTUserProvider;
```
chalasr: This use is now unused

chalasr:
Thank you @sroze

srigi:
This is merged, but not released as a tagged composer package.
Please release this, our team would much appreciated that.

PR #349, opened 2017-06-14
Moved mandatory dependency in composer.json

zebba:
When using an instance of `lexik_jwt_authentication.encoder.lcobucci` an error message shows up:

```
  [Symfony\Component\Debug\Exception\ClassNotFoundException]                  
  Attempted to load class "Sha512" from namespace "Lcobucci\JWT\Signer\Rsa".  
  Did you forget a "use" statement for another namespace?
```

This error can be triggered if you did not install the dev-dependencies by simply retrieving the service from the container:

```php
$service = $this->getContainer()->get('lexik_jwt_authentication.encoder.lcobucci');
```

Moving the dependency fixes the issue.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,8 @@
         "symfony/framework-bundle": "^2.8|^3.0",
         "symfony/security-bundle": "^2.8|^3.0",
         "symfony/console": "^2.8|^3.0",
-        "namshi/jose": "^7.2"
+        "namshi/jose": "^7.2",
+        "lcobucci/jwt": "~3.2"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.8|^5.0",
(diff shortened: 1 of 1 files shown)


chalasr:
In fact, using the lcobucci encoder is optional, it's not the default one. Making the `lcobucci/jwt` a soft dependency is made on purpose, it should remain as is until we make it our default JOSE library (which is planed for the next major). 

Meanwhile, we should provide a meaningful exception when the package is missing, e.g. in the encoder constructor. Would you mind to do it in this PR instead?

zebba:
I actually don't believe throwing an exception is the right way to solve this: it will still prevent the dependency injection container to be build entirely because creating an instance will still fail.

However I assume we can get rid of the Lcobucci-related services from within the `LexikJWTAuthenticationExtension` if the library is not present.

chalasr:
You're right, the right fix would be to not register these services if the library is missing

chalasr:
@zebba Would you mind to finish this or shall I take it over?

zebba:
I've tried to look into it, but was not able to solve it yet.

chalasr:
Fixed in #383

PR #352, opened 2017-06-16
JWT header alteration

### LINKED ISSUE #351: How to add a HEADER parameter?
Say I want to add a `"typ": "JWT"` to the header. How do I accomplish it?
Thank you!

Spomky:
This PR tries to fix #351 .

Please note that method signatures are changed.
BC is kept but a minor release seems to be needed.

* [x] Tests added
* [x] Documentation updated.

### DIFF
--- a/Encoder/DefaultEncoder.php
+++ b/Encoder/DefaultEncoder.php
@@ -11,7 +11,7 @@ use Lexik\Bundle\JWTAuthenticationBundle\Services\JWSProvider\JWSProviderInterfa
  *
  * @author Robin Chalas <robin.chalas@gmail.com>
  */
-class DefaultEncoder implements JWTEncoderInterface
+class DefaultEncoder implements HeaderAwareJWTEncoderInterface
 {
     /**
      * @var JWSProviderInterface
@@ -34,7 +34,7 @@ class DefaultEncoder implements JWTEncoderInterface
         try {
             $jws = $this->jwsProvider->create($payload, $header);
         } catch (\InvalidArgumentException $e) {
-            throw new JWTEncodeFailureException(JWTEncodeFailureException::INVALID_CONFIG, 'An error occured while trying to encode the JWT token. Please verify your configuration (private key/passphrase)', $e);
+            throw new JWTEncodeFailureException(JWTEncodeFailureException::INVALID_CONFIG, 'An error occurred while trying to encode the JWT token. Please verify your configuration (private key/passphrase)', $e);
         }
 
         if (!$jws->isSigned()) {
@@ -47,7 +47,7 @@ class DefaultEncoder implements JWTEncoderInterface
     /**
      * {@inheritdoc}
      */
-    public function decode($token, array &$header = [])
+    public function decode($token)
     {
         try {
             $jws = $this->jwsProvider->load($token);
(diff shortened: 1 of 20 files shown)


Encoder/DefaultEncoder.php
```diff
@@ -29,10 +29,10 @@ public function __construct(JWSProviderInterface $jwsProvider)
     /**
      * {@inheritdoc}
      */
-    public function encode(array $payload)
+    public function encode(array $header, array $payload)
```
chalasr: Given this class can be an extension point, I think we should keep BC. We can do it by not changing the signature right now, handling the new argument only if passed and triggering a deprecation if it is not, saying that this method tzkes a new argument that will be mandatory in 3.0.
It can be done using func_num_args()/func_get_arg(), see https://github.com/symfony/symfony/blob/3.3/src/Symfony/Component/Console/Helper/QuestionHelper.php#L104 for inspiration

chalasr: Btw, let's make $header the last arg, that will give a simpler upgrade path. I would also make it optional, so we can let the interface as is and rethink about changing it in 3.0 (no upgrade path possible for interfaces)

chalasr:
Thanks for working on this @Spomky!

Spomky:
You're welcome.
I will continue that PR soon. I a quite busy this week.

Spomky:
Hi @chalasr,

Sorry for the daly.
I updated my PR with your recommandations.

Comments are welcome.

Kind regards.

Event/JWTDecodedEvent.php
```diff
@@ -22,12 +27,22 @@ class JWTDecodedEvent extends Event
...
         $this->isValid = true;
+        $this->header  = $header;
+    }
+
+    /**
+     * @return array
+     */
+    public function getHeader()
+    {
+        return $this->payload;
```
chalasr: should be $this->payload :) I guess this is what breaks tests

Encoder/JWTEncoderInterface.php
```diff
@@ -14,21 +14,23 @@
...
      * @param array $data
+     * @param array $header
      *
      * @return string the encoded token string
      *
      * @throws JWTEncodeFailureException If an error occurred while trying to create
      *                                   the token (invalid crypto key, invalid payload...)
      */
-    public function encode(array $data);
+    public function encode(array $data, array $header = []);
```
chalasr: Adding an argument here still breaks BC, even if optional. The signature in concrete implementations must match the interface one. What we need is to introduce a `HeaderAwareEncoderInterface`, make our built-in encoders implement it and check that the passed encoder does implement it before relying on the new argument in JWTManager.

Encoder/JWTEncoderInterface.php
```diff
@@ -14,21 +14,23 @@
...
      * @param string $token
+     * @param array  $header
      *
      * @return array
      *
      * @throws JWTDecodeFailureException If an error occurred while trying to load the token
      *                                   (invalid signature, invalid crypto key, expired token...)
      */
-    public function decode($token);
+    public function decode($token, array &$header = []);
```
chalasr: Actually I would not change `decode()` as we don't allow to pass a payload here, just an immutable (stringish) token, its content should not be altered at this stage I think. But maybe you have a specific use case in mind?

Encoder/JWTEncoderInterface.php
Spomky: Agreed

Encoder/JWTEncoderInterface.php
Spomky: The idea is to find a way to return the header after the decoding process, not to modify it.
But I have no specific use case and I will remove it.

chalasr:
Thanks a lot @Spomky

PR #363, opened 2017-07-08
services.xml split

Spomky:
This PR splits the service.xml file into several files.

This will allow to ease the bundle management (e.g. when deprected services will be removed) and may help to fix bugs (e.g. namshi and lcobucci services loaded while a custom encoder/provider is used).


chalasr:
That was in the [todo list](https://github.com/lexik/LexikJWTAuthenticationBundle/projects/2#card-3346971) :) Way better like this indeed.
Thanks @Spomky!

PR #369, opened 2017-07-29
Fix HMAC support

### LINKED ISSUE #307: Sign with HMAC algo is keypair bound
I'm using the bundle with lcobucci provider, his library offers an HMAC sign feature using only a secret, the bundle instead seems to not support it requiring a private/public certificate.

in \Lexik\Bundle\JWTAuthenticationBundle\Services\JWSProvider\LcobucciJWSProvider::create:66
```
new Key($this->keyLoader->loadKey('private'), $this->keyLoader->getPassphrase());

// can be too

new Key('secret')
```

and in \Lexik\Bundle\JWTAuthenticationBundle\Services\JWSProvider\LcobucciJWSProvider:91
```
$jws->verify($this->signer, $this->keyLoader->loadKey('public')) && $jws->validate(new ValidationData());

// can be too

$jws->verify($this->signer, '--> secret <--') && $jws->validate(new ValidationData());
```

do you plan to support it ? I can try to prepare a pr if it's ok for you

### LINKED ISSUE #340: Allow passing RAW keys directly in config (from "private_key_path" to "private_key")
Changing `public_key_path`/`private_key_path` to `public_key`/`private_key`, allowing to pass the key content directly as well as its file path.
This would allow storing keys as environment variables, btw removing the need for git-ignoring these keys if stored in the project itself (fixing https://github.com/symfony/recipes/pull/89).

This also relates to one thing I'd like to get in: allow using symmetric key algorithms (shared key,   https://github.com/lexik/LexikJWTAuthenticationBundle/projects/1#card-2657179)

chalasr:
- [x] Config
- [x] Encoders

Fixes #307 
Fixes #340

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -25,12 +25,23 @@ class Configuration implements ConfigurationInterface
             ->addDefaultsIfNotSet()
             ->children()
                 ->scalarNode('private_key_path')
+                    ->setDeprecated('The "%path%.%node%" configuration key is deprecated since version 2.5. Use "%path%.secret_key" instead.')
                     ->defaultNull()
                 ->end()
                 ->scalarNode('public_key_path')
+                    ->setDeprecated('The "%path%.%node%" configuration key is deprecated since version 2.5. Use "%path%.public_key" instead.')
+                    ->defaultNull()
+                ->end()
+                ->scalarNode('public_key')
+                    ->info('The key used to sign tokens (useless for HMAC). If not set, the key will be automatically computed from the secret key.')
+                    ->defaultNull()
+                ->end()
+                ->scalarNode('secret_key')
+                    ->info('The key used to sign tokens. It can be a raw secret (for HMAC), a raw RSA/ECDSA key or the path to a file itself being plaintext or PEM.')
                     ->defaultNull()
                 ->end()
                 ->scalarNode('pass_phrase')
+                    ->info('The key passphrase (useless for HMAC)')
                     ->defaultValue('')
                 ->end()
                 ->scalarNode('token_ttl')
@@ -71,9 +82,6 @@ class Configuration implements ConfigurationInterface
         return $treeBuilder;
     }
 
-    /**
-     * @return TreeBuilder
-     */
     private function getTokenExtractorsNode()
     {
         $builder = new TreeBuilder();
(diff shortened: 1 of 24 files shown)


railto:
Any update as to when this could be implemented?

chalasr:
@railto This will be released by the end of the month, some cleanup needed.

piteur:
@chalasr - Can't wait for this to be implemented. It'll be easier to deploy with docker thanks to this PR. No more volume to store only the keys...
Thanks for your work !

DependencyInjection/Configuration.php
```diff
@@ -25,12 +25,23 @@ public function getConfigTreeBuilder()
             ->addDefaultsIfNotSet()
             ->children()
                 ->scalarNode('private_key_path')
+                    ->setDeprecated('The "%path%.%node%" configuration key is deprecated since version 2.5. Use "%path%.secret_key%" instead.')
```
Oliboy50: I think the new configuration key should contain 2 or 4 `%` symbols, not 3, right?

DependencyInjection/Configuration.php
```diff
@@ -25,12 +25,23 @@ public function getConfigTreeBuilder()
...
                     ->defaultNull()
                 ->end()
                 ->scalarNode('public_key_path')
+                    ->setDeprecated('The "%path%.%node%" configuration key is deprecated since version 2.5. Use "%path%.public_key%" instead.')
```
Oliboy50: same

DependencyInjection/Configuration.php
chalasr: good catch

MasseElch:
@chalasr What is the progress here? Do you need any help?

chalasr:
Sorry for being latish guys, this is ready and will be merged today (not now in case one wants to give it a last review).

DependencyInjection/LexikJWTAuthenticationExtension.php
```diff
@@ -52,8 +49,20 @@ public function load(array $configs, ContainerBuilder $container)
...
+            $config['secret_key'] = $config['private_key_path'];
+            $container->setParameter('lexik_jwt_authentication.private_key_path', $config['secret_key']);
+        }
+
+        if (isset($config['public_key_path'])) {
+            $config['public_key'] = $config['public_key_path'];
+            $container->setParameter('lexik_jwt_authentication.public_key_path', $config['public_key']);
+        }
+
+        if (empty($config['public_key']) && empty(['secret_key'])) {
```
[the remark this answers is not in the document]
chalasr: nice catch! thanks

PR #377, opened 2017-08-15
Method getHeader() and setHeader() is not exist

ad3n:
Remove not exist methods calls


Spomky:
these methods exist: see [Event/JWTCreatedEvent.php#L43](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Event/JWTCreatedEvent.php#L43) and [Event/JWTCreatedEvent.php#L51](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Event/JWTCreatedEvent.php#L51).

See also https://github.com/lexik/LexikJWTAuthenticationBundle/pull/352 for more information.

chalasr:
Yes, that needs to be released.

PR #381, opened 2017-08-24
Fix checkIssuedAt() to allow same-second value

alexandrebintz-axiocode:
Example case where strict comparison fails : generate a token programmatically and immediately call a route with it

### DIFF
--- a/Signature/LoadedJWS.php
+++ b/Signature/LoadedJWS.php
@@ -119,7 +119,7 @@ final class LoadedJWS
      */
     private function checkIssuedAt()
     {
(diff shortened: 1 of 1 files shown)


chalasr:
Well, the current check rejects the token if the value of the `iat` claim is in the strict future, so that it's ok if it is less or equal than current time. Am I mistaken?

alexandrebintz-axiocode:
You are right. My mistake.

However I did have a problem on a recent project where I set `iat` to `(new \DateTime())->format('U')` and it did not pass validation. I had to change to `(new \DateTime('-1 hour'))->format('U')` to make it work. Can't explain why.

Anyway thanks for the good work 👍

PR #399, opened 2017-10-21
Fix sf3.4 command autoregistration deprecation

ogizanagi:
Fixes:

> Auto-registration of the command "Lexik\Bundle\JWTAuthenticationBundle\Command\CheckConfigCommand" is deprecated since Symfony 3.4 and won't be supported in 4.0. Use PSR-4 based service discovery instead


chalasr:
Thank you @ogizanagi.

PR #412, opened 2017-11-09
Make the token type case insensitive

### LINKED ISSUE #411: "bearer" is case sensitive
It looks like the the `token_type` is case insensitive, and according to the spec, it should not be, see this blog post:
https://sgeb.io/posts/2015/05/fix-go-oauth2-case-sensitive-bearer-auth-headers/

https://github.com/lexik/LexikJWTAuthenticationBundle/blob/ddfd184c8dd2f56b0b01bc5731dc84c9c9396701/TokenExtractor/AuthorizationHeaderTokenExtractor.php#L51-L53

greg0ire:
See https://tools.ietf.org/html/rfc6749#section-5.1
Fixes #411


chalasr:
Thanks @greg0ire.

PR #413, opened 2017-11-10
Use correct yaml syntax in documentation

jongotlin:


chalasr:
Good catch, thanks @jongotlin.

PR #417, opened 2017-12-01
lexik:jwt:gen-keys command

zorn-v:
Adds command that generate keys from container parameters. 
Also it runs (test) before other tests. Current keys for testing is removed and added to .gitignore

You can use this command in flex recipe then


jonag:
I think this command is a good feature to have but can't we use PHP's openssl extension or [phpseclib](https://github.com/phpseclib/phpseclib) to generate the keys instead of relying on the fact that the users have `openssl` installed and in their path?

zorn-v:
Done

bgaleotti:
Maybe extend `Symfony\Component\Console\Command\Command::isEnabled` and add a check if `openssl` extension is enabled?

zorn-v:
Maybe this bundle will not work without openssl extension ?

chalasr:
Thanks for working on this feature. The command should indeed check that openssl is installed and use it, fallback to phpseclib if available otherwise, or fail.

zorn-v:
If I understand correct, you a willing to replace `namshi/jose` to `lcobucci/jwt` and `ext-openssl` is dependence for last one (in require section).

chalasr:
Right, let's use openssl ext blindly :+1:

zorn-v:
And half of year pass,,,,

zorn-v:
Another month...
Just mark it as deprecated, nah ?

Spomky:
The method `random_bytes` would be more appropriate (available on PHP 7.0+).

sroze:
@zorn-v I guess it's up to you to update the PR isn't it? 🤔

zorn-v:
What I need to update ?

chalasr:
Nothing for now, my bad here :) 
Sorry, I spend my free time for OSS as much as possible but I'm quite busy these times. I have a big minor to draft, I'll review this asap.
 Thanks for your understanding

Spomky:
Excuse my ignorance, but why this command to generate an encrypted key pair when `openssl genrsa -out key.pem -aes256 4096` could suffice?

zorn-v:
It uses actual paths and passphrase for keys

Command/GenKeysCommand.php
```diff
@@ -0,0 +1,77 @@
...
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $fs = new Filesystem();
+        $fs->mkdir(dirname($this->privateKey));
+        $fs->mkdir(dirname($this->publicKey));
+
+        $pKeyRes = openssl_pkey_new([
+            'digest_alg'         => 'sha512',
+            'private_key_bits'   => 4096,
+            'private_key_type'   => OPENSSL_KEYTYPE_RSA,
```
chalasr: This bundle also supports ECDSA, the command should probably throw an exception in this case (and maybe support it in the future).

composer.json
```diff
@@ -49,7 +49,8 @@
         "symfony/dom-crawler": "^2.8|^3.0|^4.0",
         "symfony/yaml": "^2.8|^3.0|^4.0",
         "friendsofphp/php-cs-fixer": "^1.1|^2.8@dev",
-        "lcobucci/jwt": "~3.2"
+        "lcobucci/jwt": "~3.2",
+        "phpunit/phpunit": ">=4.0"
```
chalasr: why is that needed? `phpunit-bridge` embeds it

phpunit.xml.dist
```diff
@@ -13,6 +13,9 @@
         >
 
     <testsuites>
+        <testsuite name="LexikJWTAuthenticationBundle Gen keys command">
+            <file>./Tests/Functional/Command/GenKeysCommandTest.php</file>
+        </testsuite>
```
chalasr: does not seem necessary, is it?

Command/GenKeysCommand.php
```diff
@@ -0,0 +1,77 @@
...
+            ->setName('lexik:jwt:gen-keys')
+            ->setDescription('Generate JWT keys')
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $fs = new Filesystem();
+        $fs->mkdir(dirname($this->privateKey));
+        $fs->mkdir(dirname($this->publicKey));
```
chalasr: The bundle now supports raw keys instead of file paths (think env var), it would be great to support writing the generated key to stdout with an option

composer.json
zorn-v: Nope https://github.com/symfony/phpunit-bridge/blob/master/composer.json

phpunit.xml.dist
zorn-v: It needed for this command test run before other tests.
It generate keys for other tests and there is no needed "in repo keys" for tests

composer.json
zorn-v: Got it. You mean `simple-phpunit`. You are right, removed then.

Command/GenKeysCommand.php
zorn-v: I don't know now how to better inject `private_key_path` and `public_key_path` as they deprecated but new `secret_key` and `public_key` not in container (as deprecated ones)

PR #433, opened 2018-01-08
Added setPayload to JWTDecodedEvent analogous to JWTCreatedEvent.

vgeyer:
This is usefull to manipulate the payload after decoding. 

Example usage could be if attributes are encrypted while the JWTCreatedEvent is triggered and should be decrypted while decoding.

### DIFF
--- a/Event/JWTDecodedEvent.php
+++ b/Event/JWTDecodedEvent.php
@@ -41,7 +41,7 @@ class JWTDecodedEvent extends Event
     /**
      * @param array $payload
      */
-    public function setPayload($payload)
+    public function setPayload(array $payload)
     {
         $this->payload = $payload;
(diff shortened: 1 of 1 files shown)


vgeyer:
Any news on this?

Event/JWTDecodedEvent.php
```diff
@@ -38,6 +38,14 @@ public function getPayload()
         return $this->payload;
     }
 
+    /**
+     * @param array $payload
+     */
+    public function setPayload($payload)
```
chalasr: array typehint is missing

vgeyer:
@chalasr pardon me. :-)

chalasr:
Thank you @vgeyer.

PR #436, opened 2018-01-09
New Feature: Roles Defined Event

mnavarrocarter:
This new event introduces a new functionality that allows to the Listeners/Subscribers to override the roles the `JWTUserToken` will have. It does it by saving state on an event and applying it to the token creation process.

This change is not API breaking, since the constructor of the event still defaults to the user roles for the token definition.

I also added the proper documentation for this change. Hope it can be merged! :)


mnavarrocarter:
Sorry about this. I'll close the pull request and open a new one when I fix it.

PR #437, opened 2018-01-10
Custom encoder documentation: encode & decode methods should properly throw exceptions

### LINKED ISSUE #431: Should JWTEncoderInterface implementing class return false or throw an exception on error?
According to documentation ([Creating your own encoder](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Resources/doc/5-encoder-service.md#create-the-encoder-class)), a class implementing the `JWTEncoderInterface` should return `false` when something goes wrong in the `decode` method.

In the other hand, the [`DefaultEncoder`](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Encoder/DefaultEncoder.php#L52) seems to always return an array and throw an exception if an error occurs.

Also, [`JWTEncoderInterface::decode`](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Encoder/JWTEncoderInterface.php)'s PHPDoc states that it returns an array (no boolean):

```php
    /**
     * @param string $token
     *
     * @return array
     *
     * @throws JWTDecodeFailureException If an error occurred while trying to load the token
     *                                   (invalid signature, invalid crypto key, expired token...)
     */
    public function decode($token);
```

What should we do?

chteuchteu:
...instead of failing silently

I also ensured that the `encode` method throws a proper `JWTEncodeFailureException` exception (just like it's done in the [default encoder](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Encoder/DefaultEncoder.php#L37)). For instance, a `\DomainException` could be thrown in the `\JWT\Authentication\JWT::sign` method if an incorrect algo is provided. Mostly to highlight the fact that generic exceptions should be caught and converted into `JWT(Encode|Decode)FailureException`s.

Closes #431

### DIFF
--- a/Resources/doc/5-encoder-service.md
+++ b/Resources/doc/5-encoder-service.md
@@ -15,6 +15,8 @@ namespace AppBundle\Encoder;
 
 use JWT\Authentication\JWT;
 use Lexik\Bundle\JWTAuthenticationBundle\Encoder\JWTEncoderInterface;
+use Lexik\Bundle\JWTAuthenticationBundle\Exception\JWTEncodeFailureException;
+use Lexik\Bundle\JWTAuthenticationBundle\Exception\JWTDecodeFailureException;
 
 /**
  * NixillaJWTEncoder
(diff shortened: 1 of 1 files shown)


chalasr:
Merged into 2.4, thank you @chteuchteu.

PR #447, opened 2018-02-02
Update index.md

janmyszkier:
do not tell people to modify dist file, it is not used by the application


chalasr:
Merged on the lowest maintained branch, thank you @janmyszkier.

PR #449, opened 2018-02-05
updated installation guide

tarach:
New version of openssl generates encrypted private keys which are unusable. I had too look for a way go decrypt it. This gave me a little headache.


chalasr:
Thank you @tarach.

PR #467, opened 2018-03-04
Fix typo in the path of jwt key directory.

hassanjuniedi:


piotaixr:
If i'm not mistaken, putting the key in the `var/` folder is for Symfony <4.
As Symfony 4 and bundle-less apps, there is a `config/` folder at the root that seems the correct way to store this kind of files. An other point here is that, to my understanding `var/` is rw whereas `config/` can/should be read-only

chalasr:
@piotaixr is right here. Thanks anyway

PR #472, opened 2018-03-08
Fixed onAuthenticationFailure handler to put exception message if needed

### LINKED ISSUE #470: Exception message not returned when using onAuthenticationFailure
Hi,

I implemented a Guard Authenticator as @chalasr  recommended #468 to do and first I want to thank him.

I also extended my User entity to handle `isEnabled()` functionality https://symfony.com/doc/current/security/entity_provider.html#forbid-inactive-users-advanceduserinterface

But now when I try to get a token (login) from a non-active user I get the following response :
```json
{
	"code": 401,
	"message": "Bad credentials"
}
```

Here is my `onAuthenticationFailure` method on my guard authenticator:
```php
public function onAuthenticationFailure(Request $request, AuthenticationException $exception)
{
    // Here $exception->getMessage() = "User account is disabled."

    return $this->failureHandler->onAuthenticationFailure($request, $exception);
}
```

Maybe `onAuthenticationFailure` function have to passed this exception message as the first parameter of `JWTAuthenticationFailureResponse`

```php
public function onAuthenticationFailure(Request $request, AuthenticationException $exception)
{
    $event = new AuthenticationFailureEvent($exception, new JWTAuthenticationFailureResponse());

    $this->dispatcher->dispatch(Events::AUTHENTICATION_FAILURE, $event);

    return $event->getResponse();
}
```

But when I'm using a JWT token from a non-active user, I get the correct response:
```json
{
	"code": 401,
	"message": "Account is disabled."
}
```

Walkoss:
Fix #470

### DIFF
--- a/Security/Http/Authentication/AuthenticationFailureHandler.php
+++ b/Security/Http/Authentication/AuthenticationFailureHandler.php
@@ -35,13 +35,7 @@ class AuthenticationFailureHandler implements AuthenticationFailureHandlerInterf
      */
     public function onAuthenticationFailure(Request $request, AuthenticationException $exception)
     {
-        if ('' != $exception->getMessage()) {
-            $response = new JWTAuthenticationFailureResponse($exception->getMessage());
-        } else {
-            $response = new JWTAuthenticationFailureResponse();
-        }
-
-        $event = new AuthenticationFailureEvent($exception, $response);
+        $event = new AuthenticationFailureEvent($exception, $exception->getMessage() ? new JWTAuthenticationFailureResponse($exception->getMessage()) : new JWTAuthenticationFailureResponse());
         $this->dispatcher->dispatch(Events::AUTHENTICATION_FAILURE, $event);
 
         return $event->getResponse();
--- a/Security/Http/Authentication/AuthenticationFailureHandler.php
+++ b/Security/Http/Authentication/AuthenticationFailureHandler.php
@@ -35,8 +35,13 @@ class AuthenticationFailureHandler implements AuthenticationFailureHandlerInterf
      */
     public function onAuthenticationFailure(Request $request, AuthenticationException $exception)
     {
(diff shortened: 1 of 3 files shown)


Tests/Functional/GetTokenTest.php
```diff
@@ -57,7 +57,7 @@ public function testGetTokenFromInvalidCredentials()
         $this->assertArrayHasKey('message', $body, 'The response should have a "message" key containing the failure reason.');
         $this->assertArrayHasKey('code', $body, 'The response should have a "code" key containing the response status code.');
 
-        $this->assertSame('Bad credentials', $body['message']);
+        $this->assertSame('Bad credentials.', $body['message']);
```
Walkoss: This is because of [that](https://github.com/symfony/security/blob/master/Core/Authentication/Provider/UserAuthenticationProvider.php#L67)
As `BadCredentialsException` extends `AuthenticationException` and set its message to `Bad credentials.` then the old assert is wrong now

Security/Http/Authentication/AuthenticationFailureHandler.php
```diff
@@ -35,8 +35,13 @@ public function __construct(EventDispatcherInterface $dispatcher)
      */
     public function onAuthenticationFailure(Request $request, AuthenticationException $exception)
     {
-        $event = new AuthenticationFailureEvent($exception, new JWTAuthenticationFailureResponse());
+        if ('' != $exception->getMessage()) {
```
piotaixr: Maybe a dumb question, but:
Any reason to this if?
Because calling the `JWTAuthenticationFailureResponse` constructor without arguments defaults to calling it with empty string which is the condition to call the constructor without argument.

chalasr:
Thank you @Walkoss.

Walkoss:
Hi @chalasr 
Do you know when 2.4.5 is planned ?
I need this feature.
Thank you

inshop:
Looks like it was missed in newer versions, @chalasr could you please merge it to the master?

PR #473, opened 2018-03-08
Update index.md

ebuildy:


chalasr:
Good catch, thanks @ebuildy.

PR #485, opened 2018-04-11
Add a `lexik:jwt:generate-token` command

sroze:
For debugging purposes (and more depending on how the tokens are used - i.e. as API keys for example) it's very useful to be able to generate tokens... therefore let's ship a command that does that 👍

### DIFF
--- a/Command/GenerateTokenCommand.php
+++ b/Command/GenerateTokenCommand.php
@@ -1,0 +1,89 @@
+<?php
+
+namespace Lexik\Bundle\JWTAuthenticationBundle\Command;
+
+use Lexik\Bundle\JWTAuthenticationBundle\Services\JWTTokenManagerInterface;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Security\Core\User\UserProviderInterface;
+
+/**
+ * GenerateTokenCommand.
+ *
+ * @author Samuel Roze <samuel.roze@gmail.com>
+ */
+class GenerateTokenCommand extends Command
+{
+    protected static $defaultName = 'lexik:jwt:generate-token';
+
+    private $tokenManager;
+
+    /** @var \Traversable|UserProviderInterface[] */
+    private $userProviders;
+
+    public function __construct(JWTTokenManagerInterface $tokenManager, \Traversable $userProviders)
+    {
+        parent::__construct();
+
+        $this->tokenManager  = $tokenManager;
+        $this->userProviders = $userProviders;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName(static::$defaultName)
+            ->setDescription('Generates a JWT token')
+            ->addArgument('username', InputArgument::REQUIRED)
+            ->addOption('user-class', 'c', InputOption::VALUE_REQUIRED)
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        if ($this->userProviders instanceof \Countable && 0 === \count($this->userProviders)) {
+            throw new \RuntimeException('You must have at least 1 configured user provider to generate a token.');
+        }
+
+        if (!$userClass = $input->getOption('user-class')) {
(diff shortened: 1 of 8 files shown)


Command/GenerateTokenCommand.php
```diff
@@ -0,0 +1,58 @@
...
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $userClass = $input->getOption('user-class');
+        $token     = $this->tokenManager->create(new $userClass($input->getArgument('username'), null));
+
+        $output->writeln([
+            '',
+            '   <info>'.$token.'</info>',
```
Oliboy50: why did you put a space before the `<info>` tag?

Command/GenerateTokenCommand.php
```diff
@@ -0,0 +1,58 @@
...
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $userClass = $input->getOption('user-class');
+        $token     = $this->tokenManager->create(new $userClass($input->getArgument('username'), null));
```
Oliboy50: my User entity does not accept any arguments in its constructor, so this will most likely fail with my specific `user-class`


FYI, when I make the same command in my projects, this is how I retrieve the token:
```php
        $user = $this->userProvider->loadUserByUsername($input->getArgument('username'));
        $token = $this->tokenManager->create($user);
```

where `userProvider` is a `UserProviderInterface` which I map to my user provider service like this:

```yaml
services:
    _defaults:
        autowire: true
    # ...
    Symfony\Component\Security\Core\User\UserProviderInterface: '@security.user.provider.concrete.entity_provider'

# note: "entity_provider" is the name of my provider in "security.yaml"
```

So, maybe this would be safer to have a `user-provider-class` option instead of `user-class` (or even configuring this to the `lexik_jwt_authentication.yaml` file)
, since we can check that the given class implements the `UserProviderInterface`

Command/GenerateTokenCommand.php
sroze: So it looks nice :)

Command/GenerateTokenCommand.php
chalasr: I agree that the current solution is not ideal, but I'm not fond of `user-provider-class` either as it implies to have a service named accordingly, whereas you could have several providers based on the same class (e.g. an `entity` provider for two different entities).

What about giving all configured user providers to the command (as a service locator, indexed by their short name) and change the option name to `user-provider`? 
In case no `--user-provider` is passed, if there is only one configured provider then use it, fail otherwise  (saying that the option must be passed since multiple providers exist).
Providers can be retrieved from the context listener definition, see https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/SecurityBundle/DependencyInjection/SecurityExtension.php#L180.
I can help if you like it

chalasr:
Definitely 👍 for adding this.

Command/GenerateTokenCommand.php
chalasr: adding extra chars would prevent redirecting the output (e.g. to clipboard), I would be very raw here :)

sroze: Meh, fair point.

sroze:
I've updated the PR to use user providers. Though, I can't see how we can easily get their short names so I've kept the service ID as name. I think that's OK as we have a default one used when only one provider is registered.

Resources/config/console.xml
```diff
@@ -9,6 +9,17 @@
             <argument type="service" id="lexik_jwt_authentication.key_loader" />
             <tag name="console.command" command="lexik:jwt:check-config" />
         </service>
+
+        <service id="lexik_jwt_authentication.user_provider_locator">
```
Oliboy50: should be explicitly `public="false"` to not lose DI performance in older symfony versions?

sroze: It can yep. Changed 👍

Command/GenerateTokenCommand.php
```diff
@@ -0,0 +1,89 @@
...
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        if ($this->userProviders instanceof \Countable && 0 === \count($this->userProviders)) {
+            throw new \RuntimeException('You must have at least 1 configured user provider for generating a token.');
```
sroze: > You must have at least one configured user provider to generate a token
?

chalasr:
Thank you @sroze.

PR #490, opened 2018-05-01
update var/ path to new config/ path

Oliboy50:
## Why?
Because https://github.com/lexik/LexikJWTAuthenticationBundle/pull/473 updated a configuration path from `var` to `config`, but forgot about all other paths

## How?
Update every occurence of `var/` to `config/`


chalasr:
Good catch, thanks @Oliboy50.

PR #492, opened 2018-05-14
[Feature] clock skew

patrickjDE:
* implement clock_skew parameter
* update unit test

This PR introduces the clock skew parameter, that allows to account for time differences between servers.


chalasr:
Could you also update the `LcobucciJWSProvider` class? I can do it otherwise.
Thanks for proposing this feature

patrickjDE:
I needed this for a project I'm working on, so why not share the implementation 😄 
I'll update the other provider class in the next days.

patrickjDE:
Finally had the time for it, also updated the corresponding tests.

chalasr:
Thanks @patrickjDE for working on this feature, this is much appreciated.

PR #493, opened 2018-05-15
index.md: Change openssl command to not ask for passphrase

blockcounter:
The -aes256 option means the key will be encrypted with aes using a passphrase,
removing this option means openssl won't ask for a passphrase and the extra step
of taking the key trough openssl again becomes unnecessary.

### DIFF
--- a/Resources/doc/index.md
+++ b/Resources/doc/index.md
@@ -32,15 +32,8 @@ Generate the SSH keys :
 
 ``` bash
 $ mkdir -p config/jwt # For Symfony3+, no need of the -p option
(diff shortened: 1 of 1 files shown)


chalasr:
I'm not sure about this change. 
Using a passphrase is a good practice in general. Even if someone reads your private key, it can't use it.

chalasr:
Well, private keys with passphrase do work on my side (see https://github.com/chalasr/lexik-jwt-authentication-sandbox). I think we should remove this sentence, it has been added recently due to issues related to last openssl versions. but it is probably not the right solution. If openssl changed meanwhile, we should rather update the code to support it.

Issue #441, opened 2018-01-17
Config token_ttl time not working


pablonunez:
Testing my project I tried to change de token_ttl value but it didn't work.

I think there is a bug on DefaultJWSProvider when create the payload. The ttl configuration is overwritten.

I fix it in this commit. Hope it's useful. 
https://github.com/pablonunez/LexikJWTAuthenticationBundle/commit/82b25ae5c1ce8cf31ff69d01c20d54a9d95099a6

MaximeMaillet:
Have you make pull request for this ?

PR #503, opened 2018-06-07
TTL does not working

MaximeMaillet:
Related in [#441](https://github.com/lexik/LexikJWTAuthenticationBundle/issues/441)


chalasr:
I'm 👍 to allow setting the exp claim (bypassing the configuration). But I think that if I set it from an event listener, I expect the value to not be changed later, I mean not adding the configured ttl.

Services/JWSProvider/DefaultJWSProvider.php
```diff
@@ -86,7 +86,7 @@ public function create(array $payload, array $header = [])
         $claims        = ['iat' => time()];
 
         if (null !== $this->ttl) {
-            $claims['exp'] = time() + $this->ttl;
+            $payload['exp'] = (isset($payload['exp']) ? $payload['exp'] : time()) + $this->ttl;
```
chalasr: I would just not touch it if it is set

chalasr:
ping @MaximeMaillet :)

Issue #523, opened 2018-07-04
Deprecation Notices using PHPUnit


Faecie:
Hey, there! Thank you for the bundle!
I stuck with the problem that unit-tests do not pass due to a deprecation notice
```1x: The "Lexik\Bundle\JWTAuthenticationBundle\Services\KeyLoader\OpenSSLKeyLoader" class is deprecated since version 2.5 and will be removed in 3.0. Use "Lexik\Bundle\JWTAuthenticationBundle\Services\KeyLoader\RawKeyLoader" instead.```

My bundle config is easy as
```
lexik_jwt_authentication:
    public_key: '%env(resolve:JWT_PUBLIC_KEY)%'
security:
  providers:
    jwt:
      lexik_jwt: ~
  firewalls:
    dev:
      pattern: ^/(_(profiler|wdt)|css|images|js)/
      security: false
    api:
      pattern:   ^/api
      stateless: true
      provider: jwt
      guard:
        authenticators:
          - lexik_jwt_authentication.jwt_token_authenticator
  access_control:
    - { path: ^/api, roles: IS_AUTHENTICATED_FULLY }
```

I xdebugged the problem and found out, that it happens because of in a process of building container Symfony tries to initiate reflection of the class OpenSSLKeyLoader that has error triggering the E_USER_DEPRECATED 
All that we need here is to declare service as deprecated in a container

```
<service id="lexik_jwt_authentication.key_loader.openssl" class="Lexik\Bundle\JWTAuthenticationBundle\Services\KeyLoader\OpenSSLKeyLoader" parent="lexik_jwt_authentication.key_loader.abstract">
            <deprecated>The "%service_id%" service is deprecated since version 2.5 and will be removed in 3.0. Use lexik_jwt_authentication.key_loader.raw instead.</deprecated>
        </service>
```

Then Symfony won't touch this service and we are good to go. Please could you fix it, or I myself could open PR

lexik/jwt-authentication-bundle:2.5.1

**Comment by Spomky** [comment]
You can set the env var `SYMFONY_DEPRECATIONS_HELPER` with value `weak` directly from your command line or with the following PHPUnit config:

```xml
<php>
    <env name="SYMFONY_DEPRECATIONS_HELPER" value="weak" />
</php>
```

chalasr:
Still a bug, you shouldn't get the notice as you're not using the service (it is also reported in the profiler panel when running the app).
👍 for the proposed solution, PR welcomed :) 
[...]

PR #525, opened 2018-07-04
Make openssl key loader service deprecated

Faecie:
Service OpenSSLKeyLoader is deprecated since version 2.5. It triggers
deprecation error as soon as someone tries to include it. Symfony
framework tries to create reflection of it while building the container
and it triggers deprecation as well. Symfony doesn't know anything about this
service being deprecated. Service should be declared as deprecated in
service config, then Symfony won't touch it.

[https://github.com/lexik/LexikJWTAuthenticationBundle/issues/523]


chalasr:
Thanks for fixing this bug @Faecie.

Faecie:
@chalasr thanks, All sorted!

PR #529, opened 2018-07-12
Update README.md

m12m3r:
Fixing Upgrade link.


chalasr:
Good catch, thanks @m12m3r.

PR #537, opened 2018-07-23
Customizable User ID Claim

Spomky:
This PR modifies the claim used to store the user ID in tokens.
Prior this PR, the claim was the same as the `user_entity_field` option (e.g. `username`).

Now the user ID claim can be customized to the standard claim `sub` (subject) as per the [RFC7519 section 4.1.2](https://tools.ietf.org/html/rfc7519#section-4.1.2) or any other value (e.g. `user_id`).

A new configuration option s added: `user_id_claim`. If null, it will have the same value as `user_entity_field`.

This PR also allow a BC with tokens that have already been issued by an application.

### DIFF
--- a/Security/Authentication/Provider/JWTProvider.php
+++ b/Security/Authentication/Provider/JWTProvider.php
@@ -45,14 +45,21 @@ class JWTProvider implements AuthenticationProviderInterface
     protected $userIdentityField;
 
     /**
+     * @var string
+     */
+    private $userIdClaim;
+
+    /**
      * @param UserProviderInterface    $userProvider
      * @param JWTManagerInterface      $jwtManager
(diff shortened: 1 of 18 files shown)


Security/Authentication/Provider/JWTProvider.php
```diff
@@ -44,22 +44,30 @@ class JWTProvider implements AuthenticationProviderInterface
      */
     protected $userIdentityField;
 
+    /**
+     * @var string
+     */
+    protected $userIdClaim;
```
chalasr: let's make it private to reduce the maintenance cost, it's exposed through a getter

Spomky:
Should be fine now.
Minor comment took into account.

chalasr:
Thank you @Spomky.

PR #542, opened 2018-08-02
Fix missing implements breaking JWT header alteration

tucksaun:


Spomky:
Excellent.Missed that none when I submitted the PR

chalasr:
Released in v2.5.4

PR #544, opened 2018-08-23
Example with Symfony > 4.0

Snowbaha:
Since SF 4.0, in the controller, it's better to use the interface instead of the service container.
Maybe it can be nice to show the 2 example ?


chalasr:
Thank you @Snowbaha.

PR #546, opened 2018-08-26
Update instructions to point to the right file

nei:
It seems routing.yml is actually not the right file, the correct is routes.yml

### DIFF
--- a/Resources/doc/index.md
+++ b/Resources/doc/index.md
@@ -85,7 +85,7 @@ security:
         - { path: ^/api,       roles: IS_AUTHENTICATED_FULLY }
 ```
 
-Configure your `routing.yml` :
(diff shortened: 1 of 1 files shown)


Resources/doc/index.md
```diff
@@ -85,7 +85,7 @@ security:
         - { path: ^/api,       roles: IS_AUTHENTICATED_FULLY }
 ```
 
-Configure your `routing.yml` :
+Configure your `config/routes.yml` :
```
chalasr: shouldn't it be `config/routes.yaml`?

nei: that is correct :) I will change it. thanks @chalasr

chalasr:
Good catch, thanks @nei.

PR #548, opened 2018-09-01
Add: Symfony4 exemple link of sandbox

alborq:
Add Symfony 4 sandbox link. 

Il have keep SF2 exemple because on openssl generation token we can see : 
`$ mkdir -p config/jwt # For Symfony3+, no need of the -p option`


chalasr:
Good catch, thanks @alborq.

PR #564, opened 2018-10-10
Add ext-openssl requirement

### LINKED ISSUE #556: openssl requirement is implicit
From documentation:

>This bundle requires Symfony 3.4+ and the openssl extension.

Anyway, openssl requirement is not present in composer.json.
It should contains "ext-openssl" in `require` section.

garak:
This should fix #556


chalasr:
Thank you @garak.

PR #570, opened 2018-10-26
Fix case

ousmaneNdiaye:


chalasr:
Good catch, thanks @ousmaneNdiaye.

Issue #576, opened 2018-11-02
JWT Provider is called with wrong parameters


astronati:
## "symfony/symfony": "^3.4.17"

```Type error: Argument 3 passed to Lexik\Bundle\JWTAuthenticationBundle\Security\Authentication\Provider\JWTProvider::__construct() must implement interface Symfony\Component\EventDispatcher\EventDispatcherInterface, string given, called in /Users/andreastronati/MagicGol/api/var/cache/dev/ContainerXgx0ht6/getSecurity_Authentication_Provider_Jwt_ApiService.php on line 10```

The error above is generated trying to access the documentation page generated by `nelmio_api_doc`

`security.yml`
```
firewalls:
        # disables authentication for assets and the profiler, adapt it according to your needs
        dev:
            pattern: ^/(_(profiler|wdt)|css|images|js)/
            security: false

        doc:
            # https://symfony.com/doc/current/security.html#a-configuring-how-your-users-will-authenticate
            pattern: ^/doc
            http_basic:
                provider: in_memory

        api_login:
            pattern: ^/v1/login
            stateless: true
            anonymous: true
            form_login:
                check_path: /v1/login
                require_previous_session: false
                username_parameter: email
                password_parameter: password
                success_handler: lexik_jwt_authentication.handler.authentication_success
                failure_handler: lexik_jwt_authentication.handler.authentication_failure

        api_registration:
            pattern: ^/v1/registration
            stateless: true
            anonymous: true

        api_reset:
            pattern: ^/v1/reset
            stateless: true
            anonymous: true

        api_social_access:
            pattern: ^/v1/social-access
            stateless: true
            anonymous: true

        api:
            pattern: ^/v1
            stateless: true
            lexik_jwt: ~
```

fjogeleit:
Its an issue in the LexikJWTAuthenticationBundle itself. I create a PR to fix this issue.

mu4ddi3:
@astronati, error comes from cached code, can you try clear cache and check again?

qlerebours:
I had the same issue, its working with latest dependency: 
` "lexik/jwt-authentication-bundle": "dev-master" `

and

`"minimum-stability": "dev",`

in composer.json

astronati:
Thanks

PR #577, opened 2018-11-04
Fix argument order in JWTProvider service declaration

fjogeleit:
Fix related to https://github.com/lexik/LexikJWTAuthenticationBundle/issues/576


chalasr:
Good catch, thanks @fjogeleit.

Issue #578, opened 2018-11-06
Unable to find the controller for path "/login_check". The route is wrongly configured.


yooouuri:
Looked into difference issues but they won't resolve my issue. Also looked into the sandbox repo, also could not find something that could help me.

The follow error occur:

> Unable to find the controller for path "/login_check". The route is wrongly configured.

config/routes.yaml
```yaml
login_check:
    path: /login_check
    methods: POST
```

config/packages/lexik_jwt_authentication.yaml
```yaml
lexik_jwt_authentication:
    secret_key: '%env(resolve:JWT_SECRET_KEY)%'
    public_key: '%env(resolve:JWT_PUBLIC_KEY)%'
    pass_phrase: '%env(JWT_PASSPHRASE)%'
```

config/packages/security.yaml
```yaml
security:
    # https://symfony.com/doc/current/security.html#where-do-users-come-from-user-providers
    providers:
        in_memory: { memory: ~ }
    firewalls:
        dev:
            pattern: ^/(_(profiler|wdt)|css|images|js)/
            security: false
        main:
            anonymous: true
        login:
            pattern:  ^/login
            stateless: true
            anonymous: true
            json_login:
                check_path: /login_check
                success_handler: lexik_jwt_authentication.handler.authentication_success
                failure_handler: lexik_jwt_authentication.handler.authentication_failure
        api:
            pattern: ^/api
            stateless: true
            anonymous: false
            guard:
                authenticators:
                    - lexik_jwt_authentication.jwt_token_authenticator

            # activate different ways to authenticate

            # http_basic: true
            # https://symfony.com/doc/current/security.html#a-configuring-how-your-users-will-authenticate

            # form_login: true
            # https://symfony.com/doc/current/security/form_login_setup.html

    # Easy way to control access for large sections of your site
    # Note: Only the *first* access control that matches will be used
    access_control:
        - { path: ^/api, roles: IS_AUTHENTICATED_FULLY }
        - { path: ^/login, roles: IS_AUTHENTICATED_ANONYMOUSLY }
        # - { path: ^/admin, roles: ROLE_ADMIN }
        # - { path: ^/profile, roles: ROLE_USER }
```

chalasr:
My guess is that you're not sending valid JSON to the `/login_check` endpoint.
Following your configuration, here is a cURL command that should work (after replacing dummy values by a real username/password couple):
[...]

PR #579, opened 2018-11-06
Update index.md

yooouuri:
Update docs so the following error should not occur anymore:

> Unable to find the controller for path _path_. The route is wrongly configured.

See #578

### DIFF
--- a/Resources/doc/index.md
+++ b/Resources/doc/index.md
@@ -58,11 +58,16 @@ lexik_jwt_authentication:
 
 Configure your `config/packages/security.yaml` :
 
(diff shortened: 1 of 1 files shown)


chalasr:
The firewall order is a symfony thing, not related to this bundle. I'm 👎 for adding it to the doc. Thanks anyway

yooouuri:
I know its not related to this bundle but if you don't setup it right it won't work...

chalasr:
Yep, I understand the intend here, but this is likely to change, and your `main` firewall is very specific to your need (that's not common).

PR #581, opened 2018-11-09
Typo in new user_id_claim configuration info

enekochan:
As above


chalasr:
Good catch, thanks @enekochan.

PR #583, opened 2018-11-11
Make exp claim optional.

### LINKED ISSUE #521: Cant override expiration / ttl
It is not possible to override the default expiration time to unlimited.

```
$encoder = $this->container->get('lexik_jwt_authentication.encoder');
$token = $encoder->encode(['email' => $data->getEmail()]);
```

With the manager it is also not possible:
```
$jwtManager = $this->container->get('lexik_jwt_authentication.jwt_manager');
$token = $jwtManager->create($newJwtUser);
```

I also tried this to remove expiration for unlimited lifetime:

```
public function onJWTCreated(JWTCreatedEvent $event)
{
    $payload        = $event->getData();
    $payload['exp'] = null;
// unset($payload['exp']); also not working

    $event->setData($payload);
}

alexsegura:
As per the [RFC](https://tools.ietf.org/html/rfc7519#section-4.1.4), the `exp` claim is optional

> The "exp" (expiration time) claim identifies the expiration time on
   or after which the JWT MUST NOT be accepted for processing.  The
   processing of the "exp" claim requires that the current date/time
   MUST be before the expiration date/time listed in the "exp" claim.
   Implementers MAY provide for some small leeway, usually no more than
   a few minutes, to account for clock skew.  Its value MUST be a number
   containing a NumericDate value.  Use of this claim is OPTIONAL.

But, there is currently no way to achieve this, as the `LoadedJWS` class expects the key `exp` to be set. 

This Pull Request allows to unset the `exp` claim, and does not enforce the `exp` key to be present.

Fixes #521

### DIFF
--- a/Signature/LoadedJWS.php
+++ b/Signature/LoadedJWS.php
@@ -114,7 +114,11 @@ final class LoadedJWS
             return;
         }
 
-        if (!isset($this->payload['exp']) || !is_numeric($this->payload['exp'])) {
+        if (!isset($this->payload['exp'])) {
+            return;
+        }
+
+        if (!is_numeric($this->payload['exp'])) {
             return $this->state = self::INVALID;
         }
 
--- a/Tests/Signature/LoadedJWSTest.php
+++ b/Tests/Signature/LoadedJWSTest.php
@@ -29,7 +29,7 @@ final class LoadedJWSTest extends TestCase
         $jws = new LoadedJWS($payload = [], true);
 
         $this->assertSame($payload, $jws->getPayload());
-        $this->assertFalse($jws->isVerified());
+        $this->assertTrue($jws->isVerified());
         $this->assertFalse($jws->isExpired());
     }
 
@@ -60,6 +60,17 @@ final class LoadedJWSTest extends TestCase
 
         $this->assertFalse($jws->isVerified());
         $this->assertTrue($jws->isExpired());
+    }
+
+    public function testVerifiedWithNoExpClaim()
+    {
+        $payload = $this->goodPayload;
+        unset($payload['exp']);
+
+        $jws = new LoadedJWS($payload, true);
+
+        $this->assertTrue($jws->isVerified());
+        $this->assertFalse($jws->isExpired());
     }
 
     public function testVerifiedWithExpiredPayloadAccountedForByClockSkew()


Spomky:
Every claim defined by the RFC is optional.
The payload `{}` is a valid payload but does it mean something? The claims set in the payload only depend on the context.

I am convinced that the absence of claim `exp` in a security token like the ones issued by this bundle will undoubtedly lead to security issues.
Obviously, I am not the maintainer of this bundle, but I really don't think such a feature adds any benefit.

alexsegura:
Sure, it is recommended to issue short-lived JWTs, together with long-lived refresh tokens, but is it the responsibility of this bundle to decide what users should do? 🙂 

I see a bunch of people around there on the Internet wanting to generate JWTs without any `exp` claim. 

In any case, by default the bundle sets an `exp` claim, so it is fine. 
You need to attach a listener to remove the `exp` claim.

Spomky:
> is it the responsibility of this bundle to decide what users should do? 🙂

This bundle does not decide what users should do. It adopts a security strategy to avoid unauthorized acess by limiting the period of validity of the tokens. 

> I see a bunch of people around there on the Internet wanting to generate JWTs without any exp claim.

The same people as [this one](https://stackoverflow.com/q/53242831/2157818)?
I am OK with immortal tokens when it is justified. The problem is that for security tokens you cannot do that without any risks.
Refresh tokens are a good way to renew the access tokens. But for SPA, the use of these refresh tokens is not a good idea at all (e.g. they are not allowed with the OAuth2 Implicit grant type for security reasons)... and a bunch of people prefer immortal tokens instead of another way to deal with that, which is worst at the end.
OpenID-Connect introduces a nice and secured way to renew the tokens using the `prompt=none` query parameter and pre-configured authorizations. It is harder to implement, but much more secured!

I will be happy to discuss more about that topic (maybe [on glitter](https://gitter.im/spomky)). Do not hesitate to contact me (I may misunderstood your reasons andchange my mind)

fd6130:
Anyways, security concern comes first. This bundle allow you to set a longer lifespan that can act as a semi-like immortal token, just set the `token_ttl` to longer seconds (like a year or more).

alexsegura:
Since I posted the issue I have more experience with JWTs and API authentication in general, and I don't want to generate tokens that never expire anymore, I now agree with @Spomky comments above 😅

The pull request can be closed by maintainers, if they think it's useless to implement.

chalasr:
I would prefer not to go this way as well. Thank you for the PR and thanks everyone for the discussion.

PR #585, opened 2018-11-14
Allow JWTTokenAuthenticator to be extended

bartekczyz:
Changed all `private` into `protected` members to make them usable inside extending classes


chalasr:
A protected property implies makes maintenance harder as it implies backward compatibility for inheritance use cases. I would need a good reason to do so for each property :)

PR #595, opened 2018-12-06
Update Readme with Symfony 4 Version of register

alborq:
Hey @chalasr 

I have add some information for Symfony 4 User.
Bundle Registring into Readme.md is unclear with Sf4.


chalasr:
Good catch, thanks @alborq.

PR #604, opened 2019-01-05
Update of the documentation, to show how to get a decoded JWT Token info from the current request

futbolsalas15:
I don't know if _index.md_ is the right file to document this, so I'm sorry if it is the wrong place. However, I really appreciate if you can document in some place, because it is a recurrent topic that some people (like me) don't find easily in Internet.

Thanks a lot!

### DIFF
--- a/Resources/doc/index.md
+++ b/Resources/doc/index.md
@@ -127,6 +127,28 @@ By default only the authorization header mode is enabled : `Authorization: Beare
 
 See [configuration reference](1-configuration-reference.md) document to enable query string parameter mode or change the header value prefix.
 
+### 3. (Optional) Get the decoded JWT Token from a Service/Controller (via Dependency Injection)
+
+If you need to get the information that have a JWT Token from a Controller or Service for some purposes (e.g.: validate role of the user), you can:
+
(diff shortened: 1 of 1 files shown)


chalasr:
Nice PR, I'll look if I can find a better place for it, but be sure it will be in.
Thanks a ton

Resources/doc/index.md
```diff
@@ -127,6 +127,28 @@ By default only the authorization header mode is enabled : `Authorization: Beare
 
 See [configuration reference](1-configuration-reference.md) document to enable query string parameter mode or change the header value prefix.
 
+### 3. (Optional) Get the decoded JWT Token from a Service/Controller (via Dependency Injection)
```
Oliboy50, line 130:
```suggestion
### 3. (Optional) Get the decoded JWT from a Service/Controller (via Dependency Injection)
```

(T is for Token)

fd6130:
@chalasr  I think i can move this to my PR #920 and it will be a new page under `Further documentation`, so this PR can be close after that. How do you think?

chalasr:
@fd6130 Sounds good to me!

fd6130:
> @fd6130 Sounds good to me!

There is an issue regarding to this PR.  Already create a new issue for this usage.

fd6130:
Forgot to mention that this PR works for Guard but not the new authentication system. Using this PR example in latest version of this bundle will cause errors, i will make a PR for the fix soon.

chalasr:
Closing in favor of #604. Thanks for raising @futbolsalas15!

PR #609, opened 2019-01-28
allow to pass payload to JWTManager directly

videni:
hi, using event to add payload to JWT is not good enough, because this design forces us to save the payload in another place.  it is best to allow to pass payload to JWTManager directly I think, we need to add a field into the payload  which is not depended on the user object, this way can let us reuse JWTManager easier.

### DIFF
--- a/Services/JWTManager.php
+++ b/Services/JWTManager.php
@@ -57,9 +57,10 @@ class JWTManager implements JWTManagerInterface, JWTTokenManagerInterface
     /**
      * {@inheritdoc}
      */
-    public function create(UserInterface $user)
+    public function create(UserInterface $user, $payload = [])
     {
-        $payload = ['roles' => $user->getRoles()];
+        $payload = array_merge(['roles' => $user->getRoles()], $payload);
+
         $this->addUserIdentityToPayload($user, $payload);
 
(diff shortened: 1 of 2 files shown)


sylfabre:
@videni I discussed this PR with @chalasr: right now the change to `JWTManagerInterface::create()` method breaks backward compatibility for anyone who implemented this interface.

This change can only happen with a major version bump, and this library follows the same rules as the Symfony framework: deprecations must be announced with notices in a new minor version.

We may go for this workaround: we'll create our own service with `JWTEncoderInterface` as a dependency, and implement the `create()` method with this new signature.

sylfabre:
@RicoLannez the solution is to turn the method into a virtual one in the interface like in https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Security/Core/Encoder/PasswordEncoderInterface.php#L21

This virtual method would get the new signature with the `array $payload = []` argument.
And the virtual method will be implemented back as a real one in the next major version.

Could you please prepare a PR with this?

chalasr:
Sorry for very late feedback @videni.
This is going to be finished in #786. Thanks for raising the topic!

PR #612, opened 2019-02-01
Fix link in table of contents for data customization

rskuipers:
Just a tiny fix for an anchor link.


chalasr:
Good catch, thanks @rskuipers.

PR #620, opened 2019-02-18
fix missing $config variable

Oliboy50:


Oliboy50:
@chalasr thank you :)

chalasr:
Thanks for fixing this bug @Oliboy50.

PR #621, opened 2019-02-19
Change to use sf4 static client

sigriffiths:
Change to use sf4 static client and pass the credentials as json


chalasr:
Thank you @sigriffiths.

PR #628, opened 2019-03-08
Update version of lcobucci/jwt to be compatible with php 7.3

alx-web:


alx-web:
3.3 is not yet released. Waiting for it to resubmit the PR

PR #641, opened 2019-04-12
Add event annotations

RikudouSage:
This annotation is used for external tools to read information about event class, it's useful for example in PhpStorm with Symfony plugin where it can automatically create the method with correct parameters, it helps speed up development a lot.


chalasr:
Thank you @RikudouSage.

PR #644, opened 2019-04-17
Fix FC/BC layer for EventDispatcher

nicolas-grekas:
Fix https://github.com/lexik/LexikJWTAuthenticationBundle/pull/637

Having Contracts 1.1 doesn't mean EventDispatcher 4.3 is installed ;)

ping @chalasr

### DIFF
--- a/Event/Event.php
+++ b/Event/Event.php
@@ -2,10 +2,10 @@
 
 namespace Lexik\Bundle\JWTAuthenticationBundle\Event;
 
-use Symfony\Component\EventDispatcher\Event as BaseEvent;
-use Symfony\Contracts\EventDispatcher\Event as ContractsBaseEvent;
+use Symfony\Component\EventDispatcher\EventDispatcher;
+use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
-if (class_exists(ContractsBaseEvent::class)) {
+if (is_subclass_of(EventDispatcher::class, EventDispatcherInterface::class)) {
     class Event extends ContractsBaseEvent
     {
     }
--- a/Event/Event.php
+++ b/Event/Event.php
@@ -2,7 +2,9 @@
 
 namespace Lexik\Bundle\JWTAuthenticationBundle\Event;
 
+use Symfony\Component\EventDispatcher\Event as BaseEvent;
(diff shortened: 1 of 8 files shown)


Event/Event.php
```diff
@@ -2,10 +2,10 @@
 
 namespace Lexik\Bundle\JWTAuthenticationBundle\Event;
 
-use Symfony\Component\EventDispatcher\Event as BaseEvent;
-use Symfony\Contracts\EventDispatcher\Event as ContractsBaseEvent;
+use Symfony\Component\EventDispatcher\EventDispatcher;
+use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
-if (class_exists(ContractsBaseEvent::class)) {
+if (is_subclass_of(EventDispatcher::class, EventDispatcherInterface::class) {
```
tucksaun: missing end `)`

nicolas-grekas: fixed

chalasr:
Good catch, thanks @nicolas-grekas.

chalasr:
Released in v2.6.3

Event/Event.php
```diff
@@ -3,9 +3,11 @@
 namespace Lexik\Bundle\JWTAuthenticationBundle\Event;
 
 use Symfony\Component\EventDispatcher\Event as BaseEvent;
+use Symfony\Component\EventDispatcher\EventDispatcher;
 use Symfony\Contracts\EventDispatcher\Event as ContractsBaseEvent;
+use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
-if (class_exists(ContractsBaseEvent::class)) {
+if (is_subclass_of(EventDispatcher::class, EventDispatcherInterface::class)) {
     class Event extends ContractsBaseEvent
```
yceruto: Type error in https://github.com/symfony/symfony/blob/fba11b4dc34e7c589d8c30d5b6a090387d52e648/src/Symfony/Component/EventDispatcher/Debug/TraceableEventDispatcher.php#L314

Reported in this repo https://github.com/lexik/LexikJWTAuthenticationBundle/issues/647 though I'm not sure if something can be improved in Symfony.

PR #650, opened 2019-05-02
Fixed AuthenticaionFailureHandler to utilize messages from custom exceptions

### LINKED ISSUE #588: UserChecker Exception Messages are lost. It is only "bad credentials" when there is any issue with credentials
I've written a UserChecker to check for active user, and deleted account. If any of these issues happen, a CustomUserMessageAuthenticationException is thrown. With that there is specific message related to the issue. But extension always give a message of "Bad Credentials" when these exceptions are thrown. I need these detail messages to be in response. Please help. 

```php
/**
 * Class UserChecker
 * @package App\Security
 *
 * Validate username/email before authentication
 * Also make sure that the user is not deleted or inactive
 */
class UserChecker implements UserCheckerInterface
{
    /**
     * @var ValidatorInterface
     */
    private $validator;

    /**
     * UserChecker constructor.
     * @param ValidatorInterface $validator
     */
    public function __construct(ValidatorInterface $validator)
    {
        $this->validator = $validator;
    }

    /**
     * @param UserInterface $user
     */
    public function checkPreAuth(UserInterface $user) : void
    {
        if (!$user instanceof User) {
            return;
        }
        $errors = $this->validator->validate($user);

        if ($errors->count()) {
            /*
             * Uses a __toString method on the $errors variable which is a
             * ConstraintViolationList object. This gives us a nice string
             * for debugging.
             */
            //$errorsString = (string) $errors;

            throw new CustomUserMessageAuthenticationException(
                $errors, [], 1
            );

        }

        // user is deleted, show a generic Account Not Found message.
        if ($user->isDeleted()) {
            //throw new AccountDeletedException('Account Deleted');

            // or to customize the message shown
            throw new CustomUserMessageAuthenticationException(
                'Your account was deleted. Sorry about that!', [],2
            );
        }
        throw new CustomUserMessageAuthenticationException(
            'Your account is not active. Sorry about that!'
        );
        if (!$user->isActive()) {
            //throw new AccountDeletedException('Account Not Active');

            // or to customize the message shown
            throw new CustomUserMessageAuthenticationException(
                'Your account is not active. Sorry about that!'
            );
        }

    }
    /**
     * @param UserInterface $user
     */
    public function checkPostAuth(UserInterface $user) : void
    {
        if (!$user instanceof User) {
            return;
        }

    }
}
```

EresDev:
Fixed #588

### DIFF
--- a/Security/Http/Authentication/AuthenticationFailureHandler.php
+++ b/Security/Http/Authentication/AuthenticationFailureHandler.php
@@ -36,7 +36,10 @@ class AuthenticationFailureHandler implements AuthenticationFailureHandlerInterf
      */
     public function onAuthenticationFailure(Request $request, AuthenticationException $exception)
     {
-        $event = new AuthenticationFailureEvent($exception, new JWTAuthenticationFailureResponse());
+        $event = new AuthenticationFailureEvent(
+            $exception,
+            new JWTAuthenticationFailureResponse($exception->getMessage())
+        );
 
         if ($this->dispatcher instanceof ContractsEventDispatcherInterface) {
             $this->dispatcher->dispatch($event, Events::AUTHENTICATION_FAILURE);
--- a/Tests/Functional/GetTokenTest.php
+++ b/Tests/Functional/GetTokenTest.php
@@ -62,7 +62,7 @@ class GetTokenTest extends TestCase
         $this->assertArrayHasKey('message', $body, 'The response should have a "message" key containing the failure reason.');
         $this->assertArrayHasKey('code', $body, 'The response should have a "code" key containing the response status code.');
 
-        $this->assertSame('Bad credentials', $body['message']);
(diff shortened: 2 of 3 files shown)


chalasr:
Thank you @EresDev.

teohhanhui:
This is wrong. `getMessage` is not safe to be displayed to the user: https://github.com/symfony/symfony/issues/18865#issuecomment-221507833

nik-lampe:
What about this? @chalasr 

After looking at the issue @teohhanhui linked to me it seems like the code should use $exception->getMessageKey() instead of $exception->getMessage() to display a message which is safe to be displayed to the user.

nik-lampe:
Great, thanks for merging.

When will be the next release?

chalasr:
@nik-lampe v2.6.5 has been tagged

PR #655, opened 2019-05-29
new lexik:jwt:generate-key command

kilip:
This pull request will add a new command for creating new secret and public pem files based on lexik_jwt configuration using **php openssl extension**. The pem files will be generated in:
*    secret key file: ```lexik_jwt_authentication.secret_key``` paramater
*    public key file: ```lexik_jwt_authentication.public_key``` parameter

The files will be generated using passphrase from ```lexik_jwt_authentication.pass_phrase``` value.

I also fix error/bug in **LexikJWTAuthenticationExtension** because container parameter ```lexik_jwt_authentication.secret_key``` and ```lexik_jwt_authentication.public_key``` never set in this Extension (please see my 7fac3062a4b6c5826a501b0b2c7176fdd41c815c commit)


chalasr:
Implemented in #817. Thank you for the work you started here @kilip

PR #664, opened 2019-06-12
Update the openssl commands to not use "genrsa" and "rsa"

teohhanhui:
`genrsa` is superseded by `genpkey`: https://www.openssl.org/docs/manmaster/man1/openssl.html

And `pkey` is the generic "public and private key management". No need to use the algorithm-specific `rsa` command.


chalasr:
Thank you @teohhanhui.

PR #675, opened 2019-07-30
Use late static binding on JWTUser

kaznovac:
use late static binding on `JWTUser` so I as a developer can extend this class (createFromPayload can now return instance of class extending this class)

https://www.php.net/manual/en/language.oop5.late-static-bindings.php


chalasr:
Good catch, thanks @kaznovac.

PR #677, opened 2019-08-15
chore/implement-against-key-loader-interface

TiMESPLiNTER:
`LcobucciJWSProvider` currently takes an instance of `RawKeyLoader` as an argument in the constructor instead of an instance of the interface `KeyLoaderInterface` which gets implemented by the `RawKeyLoader` class.


TiMESPLiNTER:
@chalasr can be merged now :)

chalasr:
It took time, but here we go, this is in now. Thank you very much @TiMESPLiNTER.

PR #679, opened 2019-08-21
fix(): Add a PreAuthenticationJWTUserTokenInterface for the final cla…

TiMESPLiNTER:
Currently `PreAuthenticationJWTUserTokenInterface` is final but does not implement an interface. So it's impossible to test with it.


chalasr:
Hey, apologies for the delay!

TiMESPLiNTER:
> What's the use case for adding an interface?

Every public method of a final class should be covered by an interface. Else if you want to mock an object you cannot fully test it.

chalasr:
I'm going to close this PR as I don't think this deserves an interface. 
Moreover, this class is not used anymore on Symfony 5.3+ with the new `jwt` authenticator.

Thank you for proposing @TiMESPLiNTER.

PR #685, opened 2019-10-15
index.md: strip leading `$` from bash commands

chteuchteu:
It allows us to easily copy-paste them without having to manually remove the leading `$`.

Note: this is open for discussion. I'm aware that blindly copy-pasting commands from the internet without understanding them can be dangerous

### DIFF
--- a/Resources/doc/index.md
+++ b/Resources/doc/index.md
@@ -42,9 +42,9 @@ return [
 #### Generate the SSH keys:
 
 ``` bash
-$ mkdir -p config/jwt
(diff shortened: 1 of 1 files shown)


EresDev:
I think $ is an indicator of shell command. It is adding extra meaning to it and is a part of convention for documenting shell commands. So it should stay.

chteuchteu:
Not a problem, I do understand!
I'll let you close this PR

chalasr:
Thanks for proposing, but I would keep them for the raisons given by @chteuchteu. Closing for now :)

Issue #684, opened 2019-10-08
Messages translation


gregurco:
Hello.

It's quite hard to provide translated error message generated by this bundle. For example message `"Bad credentials"` is not passed through translator as it is with similar message in Security Bundle of the Symfony ([link](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Security/Core/Exception/BadCredentialsException.php)).
What do you think about this change? If it's acceptable, then I can prepare PR.

Thanks, Vlad.

EresDev:
I don't think there is a need for change in the bundle. The "Bad credentials" message usually comes from Exception and I think Exception messages are for developers and not for end users. 

Maybe a decorator of `AuthenticationFailureHandler` will be the best solution among many others.

gregurco:
@EresDev I did it in this way, but for me, it's a little bit overhead to invent an decorator/listener to replace the message provided by bundle. For me, security bundle is the great example in this case, but you decide.

EresDev:
Here is one quick solution. Simple inheritance.

```php
<?php

namespace App\ThirdParty\Security\Lexik;

use Lexik\Bundle\JWTAuthenticationBundle\Event\AuthenticationFailureEvent;
use Lexik\Bundle\JWTAuthenticationBundle\Events;
use Lexik\Bundle\JWTAuthenticationBundle\Response\JWTAuthenticationFailureResponse;
use Lexik\Bundle\JWTAuthenticationBundle\Security\Http\Authentication\AuthenticationFailureHandler;
use Symfony\Component\EventDispatcher\EventDispatcherInterface;
use Symfony\Component\HttpFoundation\Request;
use Symfony\Component\Security\Core\Exception\AuthenticationException;
use Symfony\Contracts\EventDispatcher\EventDispatcherInterface as ContractsEventDispatcherInterface;
use Symfony\Contracts\Translation\TranslatorInterface;

class TranslatedAuthenticationFailureHandler extends AuthenticationFailureHandler
{
    private $translator;

    public function __construct(EventDispatcherInterface $dispatcher, TranslatorInterface $translator)
    {
        parent::__construct($dispatcher);

        $this->translator = $translator;
    }

    public function onAuthenticationFailure(Request $request, AuthenticationException $exception)
    {

//        $event = new AuthenticationFailureEvent(
//            $exception,
//            new JWTAuthenticationFailureResponse($exception->getMessage())
//        );

        $event = new AuthenticationFailureEvent(
[...]

PR #687, opened 2019-10-22
Authentication Exception Message from its key, Explicit Travis build env, and .idea to gitignore

EresDev:
There are 3 issues addressed here. I would have used different PR for each issue but there was problem with travis build env that wouldn't have produced accurate results if not combined.

1. Updated Travis build env as you can see php 5.5 build are failing. Travis has recently [changed it default build env to Xenial](https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment) and I have explicitly specified it to be "Trusty". 

2. Addresses [the concern](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/650#issuecomment-505123009) of Authentication Exception message to be retrieved by getMessage. I have changed it to getMessageKey method to meet symfony convention, help translation and prevent possible leak of sensitive info to end user. Also addresses #684

3. Adds .idea to .gitignore
Most of use PHPStorm and it creates .idea folder that is an issue while working. It should be in .gitignore to prevent its accidental addition to repository. It also helps to work on this library quickly as you don't have to deal with .idea first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ build
 vendor
 composer.lock
 composer.phar
+.idea
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: php
 
+dist: trusty
+
 matrix:
   fast_finish: true
   include:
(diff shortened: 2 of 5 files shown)


EresDev:
@chalasr Is there something wrong that you haven't merged it?

chalasr:
Thank you @EresDev.

PR #690, opened 2019-10-31
Update 2-data-customiztion.md

patrickmaynard:
Fixes typo


chalasr:
Good catch, thanks @patrickmaynard.

PR #703, opened 2020-01-02
getRawToken() method added to Token

kacperjurak:


chalasr:
Thank you @kacperjurak.

PR #725, opened 2020-03-06
Add information about HS256

ElGovanni:
Some people get confused about ssh keys (especially windows users) so I think it could be useful.


drupol:
I think this needs a rebase.

Resources/doc/1-configuration-reference.md
```diff
@@ -23,6 +23,8 @@ lexik_jwt_authentication:
 #...
 lexik_jwt_authentication:
     secret_key: yoursecret
+    encoder:
+        signature_algorithm: HS256 # optional, set it when you want to use without SSH keys
```
chalasr, line 27: I would remove the comment here, it looks confusing.

PR #737, opened 2020-04-14
Enable to keep the modified payload after decode

cedriclombardot:


chalasr:
Great first contribution! Thank you @cedriclombardot.

Issue #743, opened 2020-05-04
Extending JWTTokenAuthenticator


Slandais:
Hi,

I want to extend JWTTokenAuthenticator to change the getUser() method (should return a different Entity depending of the structure of the payload).
For all the other methods I want to use the parent methods.

So I follow this :
https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Resources/doc/6-extending-jwt-authenticator.md

use Lexik\Bundle\JWTAuthenticationBundle\Security\Guard\JWTTokenAuthenticator as BaseAuthenticator;
class CustomJWTTokenAuthenticator extends BaseAuthenticator
{
    public function __construct(
        JWTTokenManagerInterface $jwtManager,
        EventDispatcherInterface $dispatcher,
        TokenExtractorInterface $tokenExtractor,
        DataProviderService $dataProviderService
    ) {
        $this->jwtManager                    = $jwtManager;
        $this->dispatcher                    = $dispatcher;
        $this->tokenExtractor                = $tokenExtractor;
        $this->preAuthenticationTokenStorage = new TokenStorage();
        $this->dataProviderService = $dataProviderService;
    }
}

In the constructor I use one more dependency.

I have error with the private property. For axample with the getCredentials() method, if I dont copy/paste in my custom class I have this error : 
Call to a member function decode() on null
here :  if (!$payload = $this->jwtManager->decode($preAuthToken)) {
in the parent class

If I call the parent constructor in the constructor, I have error with the preAuthenticationTokenStorage set by my custome getUser() methode but read by the createAuthenticatedToken() method of the parent class.
Both methods of the parent and child class can't work together.

How do you recomend to extending your JWTTokenAuthenticator class ?

Or is it possible to change all the private methode and attributes to protected ? => I try this by modifying the JWTTokenAuthenticator class in the vendor and it works : I can juste redeclare the getUser() method.

Here my config :  

    app.jwt_token_authenticator:
        class: App\Auth\Security\Guard\Authenticator\PartnerJWTTokenAuthenticator
        parent: lexik_jwt_authentication.security.guard.jwt_token_authenticator
        autowire: true
        autoconfigure: false

(If autowire and autoconfigure are not set I have this error : "Attribute "autowire" on service "app.jwt_token_authenticator" cannot be inherited from "_defaults" when a "parent" is set. Move your child definitions to a separate file or define this attribute explicitly in"

fd6130:
> Hi,
> 
[...]

PR #744, opened 2020-05-05
#743 protected properties for JWTTokenAuthenticator

Slandais:


odoucet:
This MR is necessary to extend JWTTokenAuthenticator.

chalasr:
I'd need more insights regarding the actual use case. Making everything protected is a no-go.
Thanks for the PR.

fd6130:
> I'd need more insights regarding the actual use case. Making everything protected is a no-go.
> Thanks for the PR.

How about provide getter method for them? So they can reuse it in their custom guard authenticator (except they want to overwrite those 4 properties) .

Issue #745, opened 2020-05-05
Extenting LcobucciJWSProvider


Slandais:
Hi,

I want to extend the LcobucciJWSProvider (Lexik\Bundle\JWTAuthenticationBundle\Services\JWSProvider\LcobucciJWSProvider) in order to rewrite the verify method (I need to verify the token with different public keys depending of data in the payload).

Actually I have to totally copy/paste the LcobucciJWSProvider to create my custom provider. But I prefer to extend the LcobucciJWSProvider and rewrite only the verify() method.

With protected properties and methods (instead of privat) it coul be easier to extend the LcobucciJWSProvider.

Thanks !

Here my config : 

    lexik_jwt_authentication.jws_provider.lcobucci:
        class: App\Auth\JWSProvider\JWSProvider
        arguments:
            - '@lexik_jwt_authentication.key_loader.raw'
            - '%lexik_jwt_authentication.encoder.crypto_engine%'
            - '%lexik_jwt_authentication.encoder.signature_algorithm%'
            - '%lexik_jwt_authentication.token_ttl%'
            - '%lexik_jwt_authentication.clock_skew%'
            - '%env(resolve:JWT_PARTNERS_PUBLIC_KEY_FOLDER)%'

fd6130:
I think this is a good suggestion for extensible provider, by changing private method to protected method and provide getter method for properties.

I can make a PR for this if it is allowed to change so.

chalasr:
Sorry but I don't want to change the visibility of these methods for maintenance reasons, I'm fine with what's required to do today although it's indeed a bit cumbersome.

PR #746, opened 2020-05-05
#745 protected properties for LcobucciJWSProvider

Slandais:


chalasr:
Hi,
Making everything protected is a no-go for me as it increases the maintenance burden. 
This class is not meant to be extended. If you have a use case that cannot be covered using event listeners or higher-level decoration, please open an issue so we can discuss about it and explore the possible solutions.
Thank you for your PR.

PR #753, opened 2020-05-27
Add `set_cookies` option to store JWT in secure cookies

### LINKED ISSUE #646: Adding a Set-Cookie header if 'cookie' is enabled?
Hello,

I just started using the bundle, it works great so far, but one thing bothers me. For now, I store my JWT tokens in localstorage in my frontend application, but it is not recommended and is a security risk. 

I read online that the tokens are usually stored in httpOnly cookies. I would love to do that with this bundle, as it seems to work with cookies as well:

```
        # check token in a cookie
        cookie:
            enabled: true
            name:    BEARER
```

This looks great, but with this config enabled, the response does not feature a [Set-Cookie](https://stackoverflow.com/a/36885/3001271) header, so that makes it useless for me.

Does  it seem like a good idea to the Set-Cookie header with the httpOnly parameter? And if not, how would you suggest to use the cookies in a secured manner?

Thank you

chalasr:
Fixes #646

- [x] More tests
- [ ] Documentation

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -4,6 +4,7 @@ namespace Lexik\Bundle\JWTAuthenticationBundle\DependencyInjection;
 
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
+use Symfony\Component\HttpFoundation\Cookie;
 
 /**
  * LexikJWTAuthenticationBundle Configuration.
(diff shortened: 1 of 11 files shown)


DependencyInjection/Configuration.php
```diff
@@ -75,6 +75,19 @@ public function getConfigTreeBuilder()
                     ->info('If null, the user ID claim will have the same name as the one defined by the option "user_identity_field"')
                 ->end()
                 ->append($this->getTokenExtractorsNode())
+                ->arrayNode('set_cookie')
```
dunglas, line 78: Could it be possible to set several cookies, or to configure the name per firewall? For instance Mercure requires to set a cookie containing a JWT names `mercureAuthorization`, but the app or other subsystems could need another token stored in another cookie.

chalasr: Node renamed to `set_cookies` which now takes an arbitrary number of cookies indexed by name, accepting the cookie lifetime, path, domain and samesite attributes as options.
Adding a `firewall` option in another PR is possible, in case setting the cookie path is not enough. Thanks!

DependencyInjection/LexikJWTAuthenticationExtension.php
```diff
@@ -88,9 +90,30 @@ public function load(array $configs, ContainerBuilder $container)
...
+
+            $cookieProviders = [];
+            foreach ($config['set_cookies'] as $name => $attributes) {
+                $container
+                    ->setDefinition($id = "lexik_jwt_authentication.cookie_provider.$name", new ChildDefinition('lexik_jwt_authentication.cookie_provider'))
+                    ->replaceArgument(0, $name)
+                    ->replaceArgument(1, $attributes['lifetime'] ?: ($config['token_ttl'] ?: 0))
+                    ->replaceArgument(2, $attributes['samesite'])
+                    ->replaceArgument(3, $attributes['path'])
+                    ->replaceArgument(4, $attributes['domain']);
```
DenisCosic, line 109: Why you dont allow to set the secure above the config?
So you can set the value for secure default = true
and the Developer can change it for development

PR #755, opened 2020-05-28
Test against symfony 5.1 and fix deprecations

acrobat:
In order to fix this deprecation I had to bump the minimum supported php version to 5.6. 
I don't know what the policy is on supporting php versions, but php 5.5 is EOL for a long time now so I guess this is ok to do?

This pr now fixes also other symfony 5.1 deprecations

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ LexikJWTAuthenticationBundle
 
 This bundle provides JWT (Json Web Token) authentication for your Symfony API.
 
-It is compatible and tested with PHP 5.5, 7.2, 7.3, on Symfony 3.4 and 4.2.
+It is compatible and tested with PHP 5.6, 7.2, 7.3, on Symfony 3.4 and 4.2.
 
 Documentation
 -------------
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -9,7 +9,7 @@ LexikJWTAuthenticationBundle
 
 This bundle provides JWT (Json Web Token) authentication for your Symfony API.
 
-It is compatible and tested with PHP 5.5, 7.2, 7.3, on Symfony 3.4 and 4.2.
+It is compatible and tested with PHP 5.6, 7.2, 7.3, on Symfony 3.4 and 4.2.
```
noniagriconomie, line 12: Maybe add Symfony 5.1 if you are testing it?
(i plan to test it an on one of my 5.1 project when 5.1 is release, could do it then if not now :))

acrobat: Done!

chalasr:
Great! Thank you @acrobat.

acasademont:
Thanks a lot! Will it be possible to release a patch version soon?

PR #763, opened 2020-06-12
Updated configuration reference

RalpinMcDjango:
Added the full default config of the automatic secure cookie generation to the config reference.


chalasr:
Thank you @RalpinMcDjango.

PR #766, opened 2020-06-20
Make chain user provider more accurate

CvekCoding:
Process chain user provider more accurate


chalasr:
Good catch, thanks @CvekCoding.

PR #767, opened 2020-06-30
Added the possibility to choose if the cookie is "secure" or not

Mael-91:
I have added the possibility to choose if the cookie is "secure" or not. This allows when we work in localhost or on a development server that is not in HTTPS that the cookie is created.

I encountered this problem today and I saw that it was not available which is a pity.

See [753](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/753#discussion_r432844990)

Have a nice evening!


chalasr:
Thank you @Mael-91.

chalasr:
Nice first contrib :) Well done

Mael-91:
Thanks to you, happy to have participated in improving this bundle.

PR #769, opened 2020-08-07
Added support for composed cookies

lukacovicadam:
I added support for composed cookies. Benefits of this approach are [described here](https://medium.com/lightrail/getting-token-authentication-right-in-a-stateless-single-page-application-57d0c6474e3). I also added ability to choose if the cookie is httpOnly or not.


chalasr:
Looks interesting, thanks! 
Regarding naming, would it make sense to use the term `split` instead of `compose/composed`?

lukacovicadam:
Thanks for tip about naming. The **split** is more accurate.

chalasr:
Thank you @lukacovicadam.

PR #771, opened 2020-08-13
Update 2-data-customization.md

Romain:
Small fix in the variable used to set the payload after adding additional data.


chalasr:
Good catch, thanks @Romain.

Romain:
You're welcome! ;)

PR #774, opened 2020-09-09
fix: "expected type of parameter 'object', 'string' provided"

sebastian87:


chalasr:
Your patch drops a backward-compatibility layer for Symfony<4, which is not good as Symfony 3.4 is still supported.
Please provide more context about the error your are facing, including the output of `composer show symfony/*` and the bundle version you are using.

sebastian87:
Oh sorry, I'm currently implementing a custom AuthenticationSuccessHandler and was wondering why my IDE was throwing a fit at the code in question :smile:

PR #775, opened 2020-09-09
EventDispatcher-Wrapper for "Symfony<4"-EventDispatcher

sebastianstucke87:
@chalasr Hi, about what you said earlier about the backwards-compatibility with Symfony <4; I was wondering why don't you just add your own wrapper for such things?

(for reference: https://github.com/lexik/LexikJWTAuthenticationBundle/pull/774#issuecomment-689544025)


Security/Firewall/JWTListener.php
```diff
@@ -41,7 +41,7 @@ class JWTListener extends AbstractListener
     protected $authenticationManager;
 
     /**
-     * @var EventDispatcherInterface
+     * @var BackwardsCompatibleEventDispatcher
```
chalasr, line 44: This is a BC break given the property is protected. We can fix it by making the wrapper implement `EventDispatcherInterface` (proxying any method call to the inner dispatcher, see https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/EventDispatcher/LegacyEventDispatcherProxy.php for inspiration)

chalasr (changes requested):
Thanks for the PR @sebastianstucke87!

chalasr:
Closing due to the lack of feedback. Thanks for the PR @sebastianstucke87

PR #780, opened 2020-09-28
Add deprecation message argument to JWTFactory.php

### LINKED ISSUE #778: addConfiguration() method in deprecated JWTFactory triggering indirect deprecation notice in Symfony 5.1
`DependencyInjection\Security\Factory\JWTFactory::addConfiguration()` calls `Symfony\Component\Config\Definition\Builder\NodeDefinition::setDeprecated()` with only two arguments:

```
if (method_exists(BaseNode::class, 'getDeprecation')) {
    $deprecationArgs = ['lexik/jwt-authentication-bundle', '2.7'];
}

if (method_exists($node, 'setDeprecated')) {
    $node->setDeprecated(...$deprecationArgs);
}
```

This results in the following notice if the config is built when running PHPUnit:
> Since symfony/config 5.1: The signature of method "Symfony\Component\Config\Definition\Builder\NodeDefinition::setDeprecated()" requires 3 arguments: "string $package, string $version, string $message", not defining them is deprecated.

Somewhat related to #762.

chrBrd:
Apologies if the message isn't quite right, but figured it was best to try a PR myself rather than expecting someone else to do it :slightly_smiling_face:.

This should fix the remaining indirect deprecation notice due to the changes made to `setDeprecated()` in Symfony 5.1.

Closes #778

### DIFF
--- a/DependencyInjection/Security/Factory/JWTFactory.php
+++ b/DependencyInjection/Security/Factory/JWTFactory.php
@@ -101,7 +101,7 @@ class JWTFactory implements SecurityFactoryInterface
     {
         $deprecationArgs = [];
         if (method_exists(BaseNode::class, 'getDeprecation')) {
-            $deprecationArgs = ['lexik/jwt-authentication-bundle', '2.7'];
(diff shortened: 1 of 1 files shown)


chalasr (approved):

chalasr:
Thank you @chrBrd.

PR #786, opened 2020-10-26
Allow token creation from an existing payload

RicoLannez:
Add virtual method in JWTTokenManagerInterface to allow token creation from an existing payload

### DIFF
--- a/Services/JWTManager.php
+++ b/Services/JWTManager.php
@@ -61,17 +61,41 @@ class JWTManager implements JWTManagerInterface, JWTTokenManagerInterface
      *
      * @return string The JWT token
      */
-    public function create(UserInterface $user, array $payload = [])
+    public function create(UserInterface $user)
+    {
+        $payload = ['roles' => $user->getRoles()];
+        $this->addUserIdentityToPayload($user, $payload);
+
+        return $this->generateJwtStringAndDispatchEvents($user, $payload);
+    }
+
+    /**
+     * @param UserInterface $user
+     * @param array $payload
+     *
+     * @return string The JWT token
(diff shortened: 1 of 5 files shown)


RicoLannez:
Hi, as discussed in this [PR](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/609) it would be great to be able to create an AccessToken from an existing payload.

Services/JWTTokenManagerInterface.php
```diff
@@ -10,6 +10,8 @@
  * JWT tokens.
  *
  * @author Robin Chalas <robin.chalas@gmail.com>
```
chalasr, line 12: Please add yourself to authors, you extended it :)

Services/JWTManager.php
```diff
@@ -88,6 +88,39 @@ public function create(UserInterface $user)
         return $jwtString;
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function createFromPayload(UserInterface $user, array $payload)
+    {
+        $payload = array_merge(['roles' => $user->getRoles()], $payload);
+        $this->addUserIdentityToPayload($user, $payload);
```
chalasr, line 97: Can you reuse the duplicated logic below by moving it to a private method called from both `create()` and `createFromPayload()`?

chalasr (commented):
2 minor comments, looks good to me otherwise. Thanks for the PR @RicoLannez

chalasr (approved):
I will merge/release tonight.

chalasr:
Thank you @RicoLannez.

PR #787, opened 2020-10-26
fix day saving transition php

flaugere:
Bug fix related to https://bugs.php.net/bug.php?id=74274
DateTime format('U') doesn't handle correctly DST. time should be
prefered


chalasr (approved):

chalasr:
Thanks for fixing this bug @flaugere.

PR #790, opened 2020-11-13
Fix Symfony 5.2 getProviderKey deprecation

ogizanagi:
👋   Hi

> Since symfony/security-http 5.2: Method "Lexik\Bundle\JWTAuthenticationBundle\Security\Authentication\Token\JWTUserToken::getProviderKey()" has been deprecated, rename it to "getFirewallName()" instead.

 from https://github.com/symfony/symfony/pull/37942

### DIFF
--- a/Security/Authentication/Token/JWTUserToken.php
+++ b/Security/Authentication/Token/JWTUserToken.php
@@ -26,7 +26,7 @@ class JWTUserToken extends AbstractToken implements GuardTokenInterface
     /**
      * {@inheritdoc}
      */
-    public function __construct(array $roles = [], UserInterface $user = null, $rawToken = null, $providerKey = null)
+    public function __construct(array $roles = [], UserInterface $user = null, $rawToken = null, $firewallName = null)
     {
         parent::__construct($roles);
 
@@ -37,7 +37,7 @@ class JWTUserToken extends AbstractToken implements GuardTokenInterface
         $this->setRawToken($rawToken);
         $this->setAuthenticated(true);
 
-        $this->providerKey = $providerKey;
+        $this->providerKey = $firewallName;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


chalasr:
Thank you @ogizanagi.

chalasr:
I will add the missing travis for 5.2 and release this in the week

Security/Authentication/Token/JWTUserToken.php
```diff
@@ -37,7 +37,7 @@ public function __construct(array $roles = [], UserInterface $user = null, $rawT
         $this->setRawToken($rawToken);
         $this->setAuthenticated(true);
 
-        $this->providerKey = $providerKey;
+        $this->providerKey = $firewallName;
```
isfar, line 40: Why don't you change the property name to `$this->firewallName` as well?

Security/Authentication/Token/JWTUserToken.php
```diff
@@ -57,11 +57,19 @@ public function getCredentials()
...
+
+        return $this->getFirewallName();
+    }
+
+    /**
+     * @return string
+     */
+    public function getFirewallName()
     {
         return $this->providerKey;
```
isfar, line 74: Same as above.

Security/Authentication/Token/JWTUserToken.php
ogizanagi: Because the property is protected, so it would have been a BC break

Spawnrad:
Hi, i have the same issue and the message 

> \"Lexik\\Bundle\\JWTAuthenticationBundle\\Security\\Authentication\\Token\\JWTUserToken::getProviderKey\" method is deprecated since version 2.10 and will be removed in 3.0. Use \"Lexik\\Bundle\\JWTAuthenticationBundle\\Security\\Authentication\\Token\\JWTUserToken::getFirewallName()\" instead. at /var/www/dovozo/vendor/lexik/jwt-authentication-bundle/Security/Authentication/Token/JWTUserToken.php:64)"}

When i debug it comes from : Symfony\Component\Security\Http\Logout\LogoutUrlGenerator.php Line 147

```
            if (null !== $token && method_exists($token, 'getProviderKey')) {
                $key = $token->getProviderKey();

                if (isset($this->listeners[$key])) {
                    return $this->listeners[$key];
                }
            }
```
I don't use logout in my application and dont have a logout on security file so i don't understand why i get this message

Issue #783, opened 2020-10-13
Functional Test Bug: The Kernel should only be booted once


Lavendar77:
Please check [Functional Test resource doc](https://github.com/lexik/LexikJWTAuthenticationBundle/blame/master/Resources/doc/3-functional-testing.md#L56).

An error seems to occur during tests - `LogicException: Booting the kernel before calling "Symfony\Bundle\FrameworkBundle\Test\WebTestCase::createClient()" is not supported, the kernel should only be booted once.`

_Symfony CLI version: v4.19.2_
_Symfony/phpunit-bridge: ^5.1_

LaurieReinette:
I have exactly the same error

Warrows:
Try removing `$client = static::createClient();` line 56.

PR #802, opened 2020-11-30
Remove duplicate client from phpunit example

Warrows:
`$client = static::createClient();` is already called line 41 and should be called only once per test function.
Closes https://github.com/lexik/LexikJWTAuthenticationBundle/issues/783


chalasr:
Good catch, thanks @Warrows.

PR #817, opened 2020-12-19
Feat: add keypair generation command

bpolaszek:
Hello there,

I'd like to remove the most annoying part of the bundle installation: leverage openssl extension to generate the key pair in a one-liner 🙂 

If you like this one, with the `-n` option, we could update the flex recipe (unfamiliar with this, but I assume that's doable) so that public / private keys are generated on the fly during `composer install`.

Thoughts?
Ben

### DIFF
--- a/Command/GenerateKeyPairCommand.php
+++ b/Command/GenerateKeyPairCommand.php
@@ -1,0 +1,225 @@
+<?php
+
+namespace Lexik\Bundle\JWTAuthenticationBundle\Command;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\Filesystem\Filesystem;
+
+/**
+ * @author Beno!t POLASZEK <bpolaszek@gmail.com>
+ */
+final class GenerateKeyPairCommand extends Command
+{
+    private const ACCEPTED_ALGORITHMS = [
+        'RS256',
+        'RS384',
+        'RS512',
+        'HS256',
+        'HS384',
+        'HS512',
+        'ES256',
+        'ES384',
+        'ES512',
+    ];
+
+    protected static $defaultName = 'lexik:jwt:generate-keypair';
+
+    /**
+     * @var Filesystem
+     */
+    private $filesystem;
+
+    /**
+     * @var string
+     */
+    private $secretKey;
+
+    /**
+     * @var string
+     */
+    private $publicKey;
+
+    /**
+     * @var string|null
+     */
+    private $passphrase;
+
+    /**
+     * @var string
+     */
+    private $algorithm;
+
+    public function __construct(Filesystem $filesystem, string $secretKey, string $publicKey, ?string $passphrase, string $algorithm)
+    {
+        parent::__construct();
+        $this->filesystem = $filesystem;
+        $this->secretKey = $secretKey;
+        $this->publicKey = $publicKey;
+        $this->passphrase = $passphrase;
+        $this->algorithm = $algorithm;
+    }
+
+    protected function configure(): void
+    {
+        $this->setDescription('Generate public/private keys for use in your application.');
+        $this->addOption('dry-run', null, InputOption::VALUE_NONE, 'Do not update key files.');
+        $this->addOption('skip-if-exists', null, InputOption::VALUE_NONE, 'Do not update key files if they already exist.');
+        $this->addOption('overwrite', null, InputOption::VALUE_NONE, 'Overwrite key files if they already exist.');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+
+        if (!in_array($this->algorithm, self::ACCEPTED_ALGORITHMS, true)) {
+            $io->error(sprintf('Cannot generate key pair with the provided algorithm `%s`.', $this->algorithm));
+
+            return 1;
+        }
+
+        [$secretKey, $publicKey] = $this->generateKeyPair($this->passphrase);
+
+        if (true === $input->getOption('dry-run')) {
+            $io->success('Your keys have been generated!');
+            $io->newLine();
+            $io->writeln(sprintf('Update your private key in <info>%s</info>:', $this->secretKey));
+            $io->writeln($secretKey);
+            $io->newLine();
+            $io->writeln(sprintf('Update your public key in <info>%s</info>:', $this->publicKey));
+            $io->writeln($publicKey);
+
+            return 0;
+        }
+
+        $alreadyExists = $this->filesystem->exists($this->secretKey) || $this->filesystem->exists($this->publicKey);
+
+        if (true === $alreadyExists) {
+            try {
+                $this->handleExistingKeys($input);
+            } catch (\RuntimeException $e) {
+                if (0 === $e->getCode()) {
+                    $io->comment($e->getMessage());
+
+                    return 0;
+                }
+
+                $io->error($e->getMessage());
+
+                return 1;
+            }
+
+            if (!$io->confirm('You are about to replace your existing keys. Are you sure you wish to continue?')) {
+                $io->comment('Your action was canceled.');
+
+                return 0;
+            }
+        }
+
+        $this->filesystem->dumpFile($this->secretKey, $secretKey);
+        $this->filesystem->dumpFile($this->publicKey, $publicKey);
+
+        $io->success('Done!');
+
+        return 0;
+    }
+
+    private function handleExistingKeys(InputInterface $input): void
+    {
+        if (true === $input->getOption('skip-if-exists') && true === $input->getOption('overwrite')) {
+            throw new \RuntimeException('Both options `--skip-if-exists` and `--overwrite` cannot be combined.', 1);
+        }
+
+        if (true === $input->getOption('skip-if-exists')) {
+            throw new \RuntimeException('Your key files already exist, they won\'t be overriden.', 0);
+        }
+
+        if (false === $input->getOption('overwrite')) {
+            throw new \RuntimeException('Your keys already exist. Use the `--overwrite` option to force regeneration.', 1);
+        }
+    }
+
+    private function generateKeyPair($passphrase): array
+    {
+        $config = $this->buildOpenSSLConfiguration();
+
+        $resource = \openssl_pkey_new($config);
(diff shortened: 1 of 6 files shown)


chalasr:
Hey,

Don't worry about the failing build, it will go away with PHP 5.6 support in #816.
I'm going to think about the best way to leverage this with flex, but I'm 👍 on the idea.
Thanks for the PR.

Resources/config/console.xml
```diff
@@ -16,6 +16,14 @@
             <argument type="collection" /> <!-- user providers -->
             <tag name="console.command" command="lexik:jwt:generate-token" />
         </service>
+
+        <service id="lexik_jwt_authentication.generate_keypair_command" class="Lexik\Bundle\JWTAuthenticationBundle\Command\GenerateKeyPairCommand" public="true">
+            <argument type="service" id="filesystem" />
+            <argument type="string">%env(default::resolve:JWT_SECRET_KEY)%</argument>
+            <argument type="string">%env(default::resolve:JWT_PUBLIC_KEY)%</argument>
+            <argument type="string">%env(default::JWT_PASSPHRASE)%</argument>
```
chalasr, line 24: Those arguments should be abstract and replaced in the DI extension (LexikJWTAuthenticationExtension) by their respective config values

Command/GenerateKeyPairCommand.php
```diff
@@ -0,0 +1,116 @@
+<?php
+
+namespace Lexik\Bundle\JWTAuthenticationBundle\Command;
+
+use BenTools\Shh\ShhException;
```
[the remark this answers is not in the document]
bpolaszek: Oops

Resources/config/console.xml
bpolaszek: Well, that's what I tried at first, but I didn't figure out how to grab the private key path / the public key path / the passphrase.

If I put a `dump($config)` around line 69 of `LexikJWTAuthenticationExtension`, here's what I get:
Capture d’écran 2020-12-20 à 17 30 34

Any idea?

chalasr: Yes, that's expected :)
Env parameters (`%env()%`) are resolved at runtime (which means they can be changed at anytime, the change will take effect on your services without clearing the cache). 
For this to work, Symfony needs to replace that `%env()%` strings by what we call "env placeholders" during the compilation phase, instead of resolving their actual value as for regular parameters (because runtime values are not known yet) so that one can manipulate such arguments in their DI extensions/compiler passes.

The test you added should be enough to make sure everything works.

bpolaszek: Alright, didn't know how it works behind the scenes 👍 
I just pushed a new commit, but it looks like I still miss something. Now when I run the command I get this:
Capture d’écran 2020-12-20 à 18 18 30

Which seems to indicate that file paths are empty.

Besides, when I set arguments as `<argument type="abstract" />` in `console.xml`, I get the following error:

> Argument 2 of service "lexik_jwt_authentication.generate_keypair_command" is abstract.  
I probably miss something here as well, your help is very welcome :-)

DependencyInjection/LexikJWTAuthenticationExtension.php
```diff
@@ -120,6 +120,12 @@ public function load(array $configs, ContainerBuilder $container)
             $container
                 ->getDefinition('lexik_jwt_authentication.handler.authentication_success')
                 ->replaceArgument(2, new IteratorArgument($cookieProviders));
+
+            $container
+                ->getDefinition('lexik_jwt_authentication.generate_keypair_command')
+                ->replaceArgument(1, $config['secret_key'])
+                ->replaceArgument(2, $config['public_key'])
+                ->replaceArgument(3, $config['pass_phrase']);
```
chalasr, line 128: Try moving the calls outside of the if block they are currently wrapped in, it should go better :)
(I must admit I had to stare at it for a while...)

bpolaszek: Eagle eyes! 😄 

(Facepalming myself) 🤦 
It works, thanks a lot! 🙏

Resources/config/console.xml
```diff
@@ -16,6 +16,14 @@
             <argument type="collection" /> <!-- user providers -->
             <tag name="console.command" command="lexik:jwt:generate-token" />
         </service>
+
+        <service id="lexik_jwt_authentication.generate_keypair_command" class="Lexik\Bundle\JWTAuthenticationBundle\Command\GenerateKeyPairCommand">
+            <argument type="service" id="filesystem" />
+            <argument type="abstract" />
+            <argument type="abstract" />
+            <argument type="abstract" />
```
chalasr, line 24: you can omit the type here (`abstract` is symfony 5.x only while this bundle supports 4.4)

bpolaszek: OK, will do.

Another thing to anticipate about the Flex recipe: correct me if I'm wrong it looks like Flex has no option about running a command during the flex recipe installation (only composer post-install, meaning it would run on every `composer install`, and this is not what we want).
One way to address this would be to add a kinda `--first-run` option, which would check for the keys existence, and dismiss if they're already here. Then we would be able to add it safely as a post-install script. WDYT?

bpolaszek: Another, safer, but less developper-friendly, would just be to add a `post-install.txt` in the flex recipe to suggest the user to run this command, then we don't need that extra behavior.

chalasr: A `--skip-if-exists` option sounds good to me, in addition to a `--overwrite` to force regeneration (which cannot be combined with the first one).

bpolaszek: Hm, I don't get it. If we have an `--skip-if-exists` option, this involves that the default behavior is overwriting, isn't it? So why having an additional `--overwrite` option?

bpolaszek: Well, if that makes more sense, I now consider that keys are not supposed to exist or an error is raised, except when using `--skip-if-exists` or `--overwrite` (I think that's what you initially meant).
I added a bunch of new tests about this (plus an encryption/decryption test) and explained that behavior in the documentation.

Command/GenerateKeyPairCommand.php
```diff
@@ -0,0 +1,150 @@
...
+
+        if (false === $input->getOption('overwrite')) {
+            throw new \RuntimeException('Your keys already exist. Use the `--overwrite` option to force regeneration.', 1);
+        }
+    }
+
+    private static function generateKeyPair($passphrase)
+    {
+        $config = [
+            'digest_alg' => 'sha512',
```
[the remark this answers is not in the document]
Spomky: The following rules should be applied:

* RS256, RS384, RS512: minimum RSA key length is 2048 bytes
* ES256, ES384, ES512: curve are SEC P-256, P-384, **P-521** respectively
* EdDSA: only curve ed25519 is supported
* HS256, HS384, HS512: minimum length **shall** be 256, 384, 512 bytes respectively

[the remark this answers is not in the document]
bpolaszek: Seems legit, but this PR will take longer than expected 😄 
I'm not 100% familiar on how to translate this into openssl config args, but will do my best.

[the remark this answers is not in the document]
bpolaszek: For example, I can't find the corresponding curve for `ed25519` inside the [curve name list](https://www.php.net/manual/en/function.openssl-get-curve-names.php) - any ideas?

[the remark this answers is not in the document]
Spomky: You can gave a look at the source code of another library I wrote (https://github.com/web-token/jwt-framework/blob/v2.2/src/Component/KeyManagement/JWKFactory.php and https://github.com/web-token/jwt-framework/blob/v2.2/src/Component/Core/Util/ECKey.php#L138)

HSxxx key is just a random string.
Both RSxxx and ESxxx need openssl and the configuration is qui te simple.
EdDSA requires libsodium, but I am not sure it can be used here.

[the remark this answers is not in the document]
bpolaszek: Well, it looks like I managed to support all algorithms, except EdDSA, which is a little more particular (user will get an error in this case). I suggest not supporting EdDSA in key pair generation, as it seems a little more exotic to achieve, unless you insist.

Tests pass with all algorithms (see [latest commit](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/817/commits/6744a05a09db60d92b777c4aae874d744d297416)), except I had to comment out the encryption/decryption test, since it only seems to work with RS* algorithms:

```
9) Lexik\Bundle\JWTAuthenticationBundle\Tests\Functional\Command\GenerateKeyPairCommandTest::testItGeneratesKeyPair with data set #14 ('HS512', 'dummy')
openssl_public_encrypt(): key type not supported in this PHP build!

/Users/benoit/Sandboxes/LexikJWTAuthenticationBundle/Tests/Functional/Command/GenerateKeyPairCommandTest.php:47

10) Lexik\Bundle\JWTAuthenticationBundle\Tests\Functional\Command\GenerateKeyPairCommandTest::testItGeneratesKeyPair with data set #15 ('ES256', 'dummy')
openssl_public_encrypt(): key type not supported in this PHP build!
```

Any ideas?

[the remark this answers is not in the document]
Spomky: OpenSSL key type for ESxxx should be `OPENSSL_KEYTYPE_EC`, not `OPENSSL_KEYTYPE_DSA`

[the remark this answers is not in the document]
bpolaszek: Fixed, thanks.

chalasr (approved):
Looks good, thanks a lot.
I will try it as soon as I have some more time before merging.

bpolaszek:
Thanks @chalasr 👍

chalasr:
It took time, but here we go, this is in now. Thank you very much @bpolaszek.

bpolaszek:
No worries / no hurries, thank you for taking the time to review this 🙂

drupol:
Thanks for the feature!

fd6130:
So this command will still use openssl to generate keypair?

Command/GenerateKeyPairCommand.php
```diff
@@ -0,0 +1,225 @@
...
+     * @var string|null
+     */
+    private $passphrase;
+
+    /**
+     * @var string
+     */
+    private $algorithm;
+
+    public function __construct(Filesystem $filesystem, string $secretKey, string $publicKey, ?string $passphrase, string $algorithm)
```
gggeek, line 56: This breaks if either secretKey or publicKey are not set in the config, as the dep injection / xml svc config will pass in NULL values and not ''.

I dunno anything about this bundle, but the doc seems to state that both secretKey and publicKey are optional...

Shall I file a ticket ?

Issue #354, opened 2017-06-23
Invalid JWT Token


walkoss:
Hi,

I'm facing a problem with your bundle.
I have two providers `customer` and `provider` and use two different endpoint to login.
I have two firewalls one for `customer` and the other for `provider`

The problem I have is that I can't use JWT token. I have a `Invalid JWT Token` error.
I think the problem comes with Symfony 3.3, not sure.

Here is my `security.yml`:

```yml
security:
    encoders:
        AppBundle\Entity\Customer:
            algorithm: bcrypt
        AppBundle\Entity\Provider:
            algorithm: bcrypt

    providers:
        customer_provider:
            entity:
                class: AppBundle:Customer
                property: email
        provider_provider:
            entity:
                class: AppBundle:Provider
                property: email

    role_hierarchy:
        ROLE_CUSTOMER:  ROLE_USER
        ROLE_PROVIDER:  ROLE_USER

    firewalls:
        dev:
            pattern: ^/(_(profiler|wdt)|css|images|js)/
            security: false

        # Customer
        api_customer_login:
            pattern:  ^/customers/login
            stateless: true
            anonymous: true
            provider: customer_provider
            form_login:
                check_path:               /customers/login
                require_previous_session: false
                username_parameter:       email
                password_parameter:       password
                success_handler:          lexik_jwt_authentication.handler.authentication_success
                failure_handler:          lexik_jwt_authentication.handler.authentication_failure

        # Provider
        api_provider_login:
            pattern:  ^/providers/login
            stateless: true
            anonymous: true
            provider: provider_provider
            form_login:
                check_path:               /providers/login
                require_previous_session: false
                username_parameter:       email
                password_parameter:       password
                success_handler:          lexik_jwt_authentication.handler.authentication_success
                failure_handler:          lexik_jwt_authentication.handler.authentication_failure

        api_provider:
            pattern:  ^/providers
            stateless: true
            lexik_jwt: ~
            provider: provider_provider

        api_customer:
            pattern:  ^/customers
            stateless: true
            lexik_jwt: ~
            provider: customer_provider

    access_control:
        - { path: login$,       role: IS_AUTHENTICATED_ANONYMOUSLY }
        - { path: ^/customers,  role: ROLE_CUSTOMER }
        - { path: ^/providers,  role: ROLE_PROVIDER }
```

I can obtain a token for these two providers without problem but when I'm using it:
```json
{
    "code": 401,
[...]

PR #825, opened 2021-01-12
Invalid JWT Token description

DeveloperGeorg:
Hi there!
I met a problem, a little time ago and got exception message "Invalid JWT Token". The main problem was, that I had this error in prod only. So, I tried to find a reason.
My first thought was, that token was signed by different private keys. I wasted tons of time to find proofs, and when I find out, that this is the wrong idea, I was confused. After lots of time I find I reason: field "iat" claim identifies the future time. Exception wasn't clear, and I think, that it could be helpful to add this information to the message.

### DIFF
--- a/Encoder/LcobucciJWTEncoder.php
+++ b/Encoder/LcobucciJWTEncoder.php
@@ -56,11 +56,19 @@ class LcobucciJWTEncoder implements JWTEncoderInterface, HeaderAwareJWTEncoderIn
         }
 
         if ($jws->isInvalid()) {
-            throw new JWTDecodeFailureException(JWTDecodeFailureException::INVALID_TOKEN, 'Invalid JWT Token', null, $jws->getPayload());
+            $message = 'Invalid JWT Token';
+            if ($jws->getStateDescription() !== '') {
+                $message .= ": {$jws->getStateDescription()}";
+            }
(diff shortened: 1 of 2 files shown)


DeveloperGeorg:
Hi, @lexikteam !
Are you still working on this pack?

chalasr:
Hi @DeveloperGeorg , yes, this is maintained :)
I need some time to test a deep review to your change because I plan to revisit the design on this area. 
I'll do as soon as possible, and I can tell you that I agree with the motivation behind the patch you propose.

fd6130:
+1 to this motivation. Because "Invalid JWT Token" just too generic and hard to debug when facing configuration error.

One of the case regarding to this error message #354

darthf1:
Is it possible to include a better description for [this](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Encoder/LcobucciJWTEncoder.php#L52) as well? I got "Invalid JWT Token" on production only. But after a while I found that the caught error was something different:
```
Lcobucci\JWT\Signer\InvalidKeyProvided
It was not possible to parse your key, reason: error:0909006C:PEM routines:get_name:no start line
```

related to https://github.com/lexik/LexikJWTAuthenticationBundle/issues/354.

PR #830, opened 2021-02-03
Remove extra \@param annotation from JWTManager::create

lucasgranberg:
Symfony gives annoying deprecation notices when using DebugClassLoader becouse of this extra annotation


chalasr (approved):

chalasr:
Good catch, thanks @lucasgranberg.

Issue #831, opened 2021-02-06
Help on how to provide a new KeyLoader


drupol:
Hello,

I'm working on a custom bundle that I will most probably release soon.

Basically the idea of this bundle is to provide a new type of [KeyLoader](https://github.com/lexik/LexikJWTAuthenticationBundle/tree/dc472dd55d6a7bc374306e3df8e6202bc95f76e9/Services/KeyLoader).
That new `KeyLoader` would make a request to fetch a public key and I also need to implements a "failsafe" mechanism when HTTP calls fails. The failsafe mechanism would then retrieve a local version of the key when the HTTP call fails.

This bundle will let you use a key defined remotely in a `jwks.json` file, eg:

```
lexik_jwt_authentication:
    secret_key: null
    public_key: 'http://something.com/jwks.json'
    pass_phrase: null
```

In order to do that, I created a `JWKSKeyLoader.php` service which is registered as such in my bundle:

```php
    $container
        ->services()
        ->set('foo.jwks_keyloader', JWKSKeyLoader::class)
        ->decorate('lexik_jwt_authentication.key_loader.raw')
        ->args([
            null, // Placeholder for $signingKey
            null, // Placeholder for $publickKey
            null, // Placeholder for $passPhrase
            service('.inner'),
            service('http_client')
        ])
        ->autowire(true)
        ->autoconfigure(true);
```

Everything so far works great.
However, when I inspect the values in `JWKSKeyLoader::__construct()`, I noticed that the `$publicKey` is always empty.

I also noticed that those values are injected in [LexikJWTAuthenticationExtension.php](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/dc472dd55d6a7bc374306e3df8e6202bc95f76e9/DependencyInjection/LexikJWTAuthenticationExtension.php#L89):

```php
        $container
            ->findDefinition('lexik_jwt_authentication.key_loader')
            ->replaceArgument(0, $config['secret_key'])
            ->replaceArgument(1, $config['public_key']);
```

Then my question is, how do you create a working decorator when its arguments are replaced? What would be the proper way to do that? Is there maybe a better way to achieve what I'm trying to do?

Thanks!

Spomky:
Hi @drupol ,

[...]

PR #832, opened 2021-02-08
Make AbstractKeyLoader::getSigningKey() and AbstractKeyLoader::getPublicKey public.

drupol:
This issue:

- [x] Make `AbstractKeyLoader::getSigningKey()` public instead of `protected`.
- [x] Make `AbstractKeyLoader::getPublicKey()` public instead of `protected`.
- [x] Update the interface `KeyLoaderInterface` with `@method` annotations.

Making them public instead of protected make possible the use of the very useful Decorator pattern.

Useful when one wants to replace one specific part of a bundle.

This PR relates to #831 but I would like to receive some feedback on this before merging.

### DIFF
--- a/Services/KeyLoader/AbstractKeyLoader.php
+++ b/Services/KeyLoader/AbstractKeyLoader.php
@@ -46,12 +46,12 @@ abstract class AbstractKeyLoader implements KeyLoaderInterface
         return $this->passphrase;
     }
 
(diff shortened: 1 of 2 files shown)


chalasr:
Thanks for your work on this new feature!

drupol:
You're welcome, thanks for your reactivity !

PR #833, opened 2021-02-08
KeyLoaderInterface::getPassphrase() might return null and we need a string.

drupol:


chalasr (approved):

chalasr:
Good catch, thanks @drupol.

drupol:
Thanks once again!

PR #835, opened 2021-02-10
Fixes #834

### LINKED ISSUE #834: bug: upgrading to 2.11.0 breaks DefaultJWSProvider (namshi/jose package missing)
Hi,

Just experienced an issue with the library.

My **lexik_jwt_authentication.yaml** looks like this:
```yaml
lexik_jwt_authentication:
   # other stuff here ...
    encoder:
        service: lexik_jwt_authentication.encoder.default
        signature_algorithm: RS256
```
As you can see it depends on **DefaultJWSProvider** and the **RS256** signature algorithm.

When upgrading to **2.11.0**, composer removes `namshi/jose` package:
```
Package operations: 0 installs, 1 update, 1 removal
  - Removing namshi/jose (7.2.3)
  - Updating lexik/jwt-authentication-bundle (v2.10.6 => v2.11.0)
```

But **DefaultJWSProvider** actually uses it (e.g., **isAlgorithmSupportedForEngine()**), so the first thing anyone will realise is an exception:
```
InvalidArgumentException: The algorithm "RS256" is not supported for OpenSSL
```

Installing the package again fixes the issue.

The thing is, DefaultJWSProvider has the `@deprecated` tag, but it says it will be supported until version 3.0, so I guess it makes sense to keep the `namshi/jose` package and deprecate both in 3.0?
```
@deprecated since version 2.5, to be removed in 3.0
```

It seems the dependency was moved to `require-dev` in #816, so It doesn't seem to have been done in purpose? Also to note, the unit tests are passing of course because at the time of running them the dependency is included (dev).

image

Thanks!

filisko:
Hi,

This PR fixes #834.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,6 +37,7 @@
         }
     ],
     "require": {
+        "namshi/jose": "^7.2",
         "php": ">=7.1",
         "ext-openssl": "*",
         "lcobucci/jwt": "^3.4|^4.0",
@@ -44,7 +45,6 @@
         "symfony/security-bundle": "^4.4|^5.1"
     },
     "require-dev": {
-        "namshi/jose": "^7.2",
         "symfony/browser-kit": "^4.4|^5.1",
         "symfony/console": "^4.4|^5.1",
         "symfony/dom-crawler": "^4.4|^5.1",


chalasr (approved):

filisko:
@chalasr you welcome, thanks for the quick response ;)

chalasr:
Released in v2.11.1 :)

wiese:
Stumbled upon this while researching why, when running `composer require jwt-auth`, a deprecated library (namshi/jose) and with it a surprising polyfill (symfony/polyfill-php56 on a php 8 system) were added.

Looking at the change which initially removed namshi/jose from here (#508), it appears to have done what it did with thought and intentionally (["The "namshi/jose" library is deprecated, this bundle does not require it anymore. **If you need to keep using it, require it in your composer.json.**"](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/508/files#diff-17a1670b7e85eaacfc2d73f538f06e48f60f2b11efbdd7df25e01afbc6a518d1R20)). In hindsight the better versioning solution might have been to release a v3 with the removal right away but that ship has sailed.

In the mean time, until there actually is a v3, "fix" #835 now puts all new users of this bundle (who care about their application supply chain) in a tight spot instead of only expecting users updating their bundle to become active in case they still depend on the deprecated library (and they really should become active anyway).

Does this resonate? Am I missing something? Are there arguments against releasing a v3 (I see one has been planned for a while but there is really no shortage of integers for version numbers)?

```
composer info lexik/jwt-authentication-bundle
...
versions : * v2.13.0
...
requires
namshi/jose ^7.2

composer why namshi/jose
lexik/jwt-authentication-bundle  v2.13.0  requires  namshi/jose (^7.2)

composer info namshi/jose
...
versions : * 7.2.3
...
requires
symfony/polyfill-php56 ^1.0

composer why symfony/polyfill-php56
namshi/jose  7.2.3  requires  symfony/polyfill-php56 (^1.0)
```

PR #840, opened 2021-02-15
[Security] On Authentication failure, replace MessageData

mpiot:
When Authentication fail, the returned message contains: `%var%` that is not replaced. 
Eg with login_throttling:

> "Too many failed login attempts, please try again in %minutes% minute."
> no replace %minutes%

In symfony they use the translator or the `strtr()` function if the translator is not used, in the case of an API login, I'm not sure the Translation is needed, then I directly use the `strtr()` function.


chalasr:
Thank you @mpiot.

PR #852, opened 2021-03-18
Add missing $payload argument in createFromPayload example

JeremieSamson:
I was reading the doc and I think there is a little something missing there.


chalasr:
Good catch, thanks @JeremieSamson.

PR #864, opened 2021-05-15
Remove development files from releases

phansys:
Environment specific files were also removed from `.gitattributes`. They should be [globally ignored](https://docs.github.com/es/github/getting-started-with-github/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer) in the environment they exist.


chalasr (approved):
👍

chalasr:
Thank you @phansys.

PR #872, opened 2021-06-08
Add new `jwt` authenticator for Symfony 5.3+ Security system

### LINKED ISSUE #810: Is it compatible with Symfony new Authenticator-based Security?
Hi, just asking if this bundle compatible with Symfony 5.2 new authenticator?

After i enable the new feature and remove the `annonymous` key from `security.yaml`, when i try to login it will use `JsonLoginAuthenticator` instead of the authenticator from this bundle.

### LINKED ISSUE #866: Using Symfony\Component\Security\Guard\AbstractGuardAuthenticator is deprecated in Symfony 5.3
Symfony 5.3 triggers this deprecation warning:

> User Deprecated: The "Lexik\Bundle\JWTAuthenticationBundle\Security\Guard\JWTTokenAuthenticator" class extends "Symfony\Component\Security\Guard\AbstractGuardAuthenticator" that is deprecated since Symfony 5.3, use the new authenticator system instead.

### LINKED ISSUE #874: 401 Unauthorized | invalid credentials | loadUserByIdentifier
Hi,

After upgrading to Sf 5.3, I get a 401 right after login.

In `vendor/lexik/jwt-authentication-bundle/Security/Guard/JWTTokenAuthenticator.php`

Replacing `return $userProvider->loadUserByUsername($identity);` with `return $userProvider->loadUserByIdentifier($identity);`
solves the problem.

Is there an update about to be published?

TristanPouliquen:
Working on implementing the new Authenticator system with Passports & Badges in LexikJWT.
Fixes #874
Fixes #866
Fixes #810

### DIFF
--- a/Resources/config/token_authenticator.xml
+++ b/Resources/config/token_authenticator.xml
@@ -15,5 +15,13 @@
                 <service class="Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorage" />
             </argument>
         </service>
+
+        <service id="lexik_jwt_authentication.security.jwt_authenticator" class="Lexik\Bundle\JWTAuthenticationBundle\Security\Authenticator\JWTAuthenticator" abstract="true">
+            <argument type="service" id="lexik_jwt_authentication.jwt_manager"/>
+            <argument type="service" id="event_dispatcher"/>
+            <argument type="service" id="lexik_jwt_authentication.extractor.chain_extractor"/>
+            <argument /> <!-- User Provider -->
+        </service>
+
     </services>
 </container>
--- a/Security/Authenticator/JWTTokenAuthenticator.php
+++ b/Security/Authenticator/JWTTokenAuthenticator.php
@@ -1,0 +1,205 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Lexik\Bundle\JWTAuthenticationBundle\Security\Authenticator;
+
+use Lexik\Bundle\JWTAuthenticationBundle\Event\JWTExpiredEvent;
+use Lexik\Bundle\JWTAuthenticationBundle\Event\JWTInvalidEvent;
+use Lexik\Bundle\JWTAuthenticationBundle\Event\JWTNotFoundEvent;
+use Lexik\Bundle\JWTAuthenticationBundle\Events;
+use Lexik\Bundle\JWTAuthenticationBundle\Exception\ExpiredTokenException;
+use Lexik\Bundle\JWTAuthenticationBundle\Exception\InvalidPayloadException;
+use Lexik\Bundle\JWTAuthenticationBundle\Exception\InvalidTokenException;
+use Lexik\Bundle\JWTAuthenticationBundle\Exception\JWTDecodeFailureException;
+use Lexik\Bundle\JWTAuthenticationBundle\Exception\MissingTokenException;
+use Lexik\Bundle\JWTAuthenticationBundle\Response\JWTAuthenticationFailureResponse;
+use Lexik\Bundle\JWTAuthenticationBundle\Security\Authentication\Token\PreAuthenticationJWTUserToken;
+use Lexik\Bundle\JWTAuthenticationBundle\Security\User\PayloadAwareUserProviderInterface;
+use Lexik\Bundle\JWTAuthenticationBundle\Services\JWTTokenManagerInterface;
+use Lexik\Bundle\JWTAuthenticationBundle\TokenExtractor\TokenExtractorInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
+use Symfony\Component\Security\Core\Exception\AuthenticationException;
+use Symfony\Component\Security\Core\Exception\UsernameNotFoundException;
+use Symfony\Component\Security\Core\Exception\UserNotFoundException;
+use Symfony\Component\Security\Core\User\ChainUserProvider;
+use Symfony\Component\Security\Core\User\UserInterface;
+use Symfony\Component\Security\Core\User\UserProviderInterface;
+use Symfony\Component\Security\Http\Authenticator\AbstractAuthenticator;
+use Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge;
+use Symfony\Component\Security\Http\Authenticator\Passport\PassportInterface;
+use Symfony\Component\Security\Http\Authenticator\Passport\SelfValidatingPassport;
+use Symfony\Component\Security\Http\EntryPoint\AuthenticationEntryPointInterface;
+use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
+
+class JWTTokenAuthenticator extends AbstractAuthenticator implements AuthenticationEntryPointInterface
+{
+    /**
+     * @var TokenExtractorInterface
+     */
+    private  $tokenExtractor;
+
+    /**
+     * @var JWTTokenManagerInterface
+     */
+    private $jwtManager;
+
+    /**
+     * @var EventDispatcherInterface
+     */
+    private $eventDispatcher;
+
+    /**
+     * @var UserProviderInterface
+     */
+    private $userProvider;
+
+    public function __construct(
+        JWTTokenManagerInterface $jwtManager,
+        EventDispatcherInterface $eventDispatcher,
+        TokenExtractorInterface $tokenExtractor,
+        UserProviderInterface $userProvider
+    ) {
+        $this->tokenExtractor = $tokenExtractor;
+        $this->jwtManager = $jwtManager;
+        $this->eventDispatcher = $eventDispatcher;
+        $this->userProvider = $userProvider;
+    }
+
+    /**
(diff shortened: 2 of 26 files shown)


TristanPouliquen:
Added unit tests to my new authenticator.

The only point that frustrates me is being forced to inject the UserProvider & store the payload as a property of the class to be able to pass them down to the `loadUser` method that calls the `loadXXXAndPayload` methods.

I didn't find a way to work around this issue and rely simply on the UserProvider structure & pass only the identifier to the UserBadge to let the magic work it all!

If anyone has an idea, I'm open to it :)

chalasr:
@TristanPouliquen Thank you very much for this PR! You just made my day :)
I can't review it right now but I'll be back at the very end of the week to hopefully help moving forward as quickly as possible.

TristanPouliquen:
@chalasr Np! Still fighting with getting it to pass all the different CIs & make it backwards compatible though

mbabker:
> The only point that frustrates me is being forced to inject the UserProvider & store the payload as a property of the class to be able to pass them down to the `loadUser` method that calls the `loadXXXAndPayload` methods.
> 
> I didn't find a way to work around this issue and rely simply on the UserProvider structure & pass only the identifier to the UserBadge to let the magic work it all!
> 
> If anyone has an idea, I'm open to it :)

I bounced this PR around in one of my apps yesterday to see if there was a way to just use Symfony's authenticator manager without the authenticator needing to be aware of the user provider, and I couldn't come up with anything.  It's basically stuck like this thanks to `PayloadAwareUserProviderInterface` and the authenticator manager not passing any information from the passport into the user loader.

Security/Authenticator/JWTTokenAuthenticator.php
```diff
@@ -0,0 +1,190 @@
...
+    {
+        return false !== $this->getTokenExtractor()->extract($request);
+    }
+
+    public function authenticate(Request $request): PassportInterface
+    {
+        $tokenExtractor = $this->getTokenExtractor();
+
+        try {
+            if (!$this->payload = $this->jwtManager->decodeFromJsonWebToken($tokenExtractor->extract($request))) {
```
mbabker, line 98:
```suggestion
            if (!$payload = $this->jwtManager->decodeFromJsonWebToken($tokenExtractor->extract($request))) {
```

This can be a local variable that's just used in the user loader later.

Security/Authenticator/JWTTokenAuthenticator.php
```diff
@@ -0,0 +1,190 @@
...
+
+        $idClaim = $this->jwtManager->getUserIdClaim();
+        if (!isset($payload[$idClaim])) {
+            throw new InvalidPayloadException($idClaim);
+        }
+
+        return new SelfValidatingPassport(
+            new UserBadge($payload[$idClaim],
+            function ($userIdentifier) {
+                return $this->loadUser($this->userProvider, $this->payload, $userIdentifier);
```
mbabker, line 117:
```suggestion
            function ($userIdentifier) use ($payload) {
                return $this->loadUser($payload, $userIdentifier);
```

Use the local payload to avoid the class var, and no need to pass the user provider into the `loadUser()` method.

Security/Authenticator/JWTTokenAuthenticator.php
```diff
@@ -0,0 +1,190 @@
...
+                }
+            }
+
+            $ex = new UserNotFoundException(sprintf('There is no user with identifier "%s".', $identity));
+            $ex->setUserIdentifier($identity);
+            throw $ex;
+        }
+
+        return $userProvider->loadUserByIdentifier($identity);
+    }
```
mbabker, line 189:
```suggestion
    /**
     * Loads the user to authenticate.
     *
     * @param array  $payload      The token payload
     * @param string $identity     The key from which to retrieve the user "identifier"
     */
    protected function loadUser(array $payload, string $identity): UserInterface
    {
        if ($this->userProvider instanceof PayloadAwareUserProviderInterface) {
            return $this->userProvider->loadUserByIdentifierAndPayload($identity, $payload);
        }

        if ($this->userProvider instanceof ChainUserProvider) {
            foreach ($this->userProvider->getProviders() as $provider) {
                try {
                    if ($provider instanceof PayloadAwareUserProviderInterface) {
                        return $provider->loadUserByIdentifierAndPayload($identity, $payload);
                    }

                    return $provider->loadUserByIdentifier($identity);
                } catch (UserNotFoundException $e) {
                    // try next one
                }
            }

            $ex = new UserNotFoundException(sprintf('There is no user with identifier "%s".', $identity));
            $ex->setUserIdentifier($identity);
            throw $ex;
        }

        return $this->userProvider->loadUserByIdentifier($identity);
    }
```

Using the user provider from the class var.

mbabker:
> make it backwards compatible though

One of the biggest things that'll help will be to check if `loadUserByIdentifier` exists on the provider and fall back to the legacy `loadUserByUsername`.  Similar to [this check](https://github.com/symfony/security-http/blob/195f7207dec4519ca8b0e407ba921ec239aa5cee/EventListener/UserProviderListener.php#L48-L55) in the `UserProviderListener`.

Security/Authenticator/JWTTokenAuthenticator.php
TristanPouliquen: When I did this earlier, I had a type error from PHPStorm, but trying it again today it doesn't pop up so who knows 🤷🏽 

It sure is cleaner so if it works 😁

TristanPouliquen:
Tests are passing on all scenarii except for PHP8 & SF5.x-dev => composer does not seem to be able to resolve a correct set of dependencies.

Someone has any idea of why?

TristanPouliquen:
> Tests are passing on all scenarii except for PHP8 & SF5.x-dev => composer does not seem to be able to resolve a correct set of dependencies.
> 
> Someone has any idea of why?

In the meantime, I've switched it to a more reasonable `5.3.x-dev` to test against the latest SF5.3

chalasr:
Sounds good 👍

mbabker:
> Someone has any idea of why?

There isn’t a 5.x branch anymore, so it can always use a version string and work right.

TristanPouliquen:
What's the next step on this PR now? :)

chalasr (changes requested):
Nice start! This misses registering the authenticator as a service (via a factory). I'm going to address my own comments and add what's missing if you don't mind.

Resources/config/token_authenticator.xml
```diff
@@ -15,5 +15,13 @@
                 <service class="Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorage" />
             </argument>
         </service>
+
+        <service id="lexik_jwt_authentication.security.jwt_authenticator" class="Lexik\Bundle\JWTAuthenticationBundle\Security\Authenticator\JWTaAuthenticator" abstract="true">
```
mbabker, line 19: Typo in the class name

```suggestion
        <service id="lexik_jwt_authentication.security.jwt_authenticator" class="Lexik\Bundle\JWTAuthenticationBundle\Security\Authenticator\JWTAuthenticator" abstract="true">
```

chalasr: Good catch

chalasr:
Now green and deprecation-free! (well, one remaining notice but purely testing/internal, should not impact developers).

Here is how this PR should be used.

Before:
```yaml
# config/packages/security.yaml
security:
   # ...
    firewalls:
        api:
            pattern:  ^/api
            stateless: true
            guard:
                authenticators: 
                    - lexik_jwt_authentication.jwt_token_authenticator
```

After:
```yaml
# config/packages/security.yaml
security:
    enable_authenticator_manager: true
    # ...
    firewalls:
        api:
            pattern:  ^/api
            stateless: true
            jwt: ~ 
```

We might add more configuration under the `jwt` key, for now there is not.
Btw, the documentation needs an update , PR welcome and much appreciated!

I'm going to merge this and draft a new release later today.

chalasr (approved):

chalasr:
Thank you very much @TristanPouliquen, this is much appreciated.

chalasr:
Released in v2.12.0.

mnocon:
Hi!

Please correct me if I'm wrong, but looks like the idea from this comment: https://github.com/lexik/LexikJWTAuthenticationBundle/pull/872#discussion_r654241854 was never executed?

There's a new method in the interface in 2.12:
https://github.com/lexik/LexikJWTAuthenticationBundle/blob/v2.12.0/Services/JWTTokenManagerInterface.php#L31-L35

Our CI setup started failing with:
```
!!  PHP Fatal error:  Class XYZ\TokenManager contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (Lexik\Bundle\JWTAuthenticationBundle\Services\JWTTokenManagerInterface::decodeFromJsonWebToken) in /var/www/vendor/XYZ/src/lib/Security/EditorialMode/TokenManager.php on line 24
```

which pointed me to this PR.

chalasr:
@mnocon You're right, I'm going to revert that breaking change this weekend. Thanks for noticing!

mnocon:
@chalasr thank you for confirming this, for now we will stick with 2.11.x and update when the next release is available.

Also thank you for maintaining this bundle! 💪

PR #878, opened 2021-06-17
Handle misc. Symfony 5.3 deprecations, update CI config

mbabker:
This PR takes a stab at handling some of the Security component deprecations in Symfony 5.3 by adding in the relevant compat checks to use the newer APIs when available and falling back to the deprecated APIs when not.  I've also cleared out some compat code that was there to support Symfony <4.4 and is no longer needed.  Lastly, a CI build for Symfony 5.3 is added in and the 5.x-dev build adjusted to target the 5.4 branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,8 +34,37 @@ jobs:
 
             - name: "Run PHPUnit Tests"
               run: "composer test"
+    php-72:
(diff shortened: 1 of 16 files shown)


chalasr (approved):

chalasr:
Thank you @mbabker.

PR #880, opened 2021-06-23
Save a little bit of memory in case the user is already in cache

mradhi:
What if we move this statement:
```php
$class = $this->class;
```

AFTER

```php
if (isset($this->cache[...])) {
       return $this->cache[...];
}
```
To save a little bit of memory in case the user is already in cache.


chalasr:
Thank you @mradhi.

PR #882, opened 2021-06-25
Fix Changelog dates for releases 2.10.6 <=> 2.12.0

tenfourzero:
`CHANGELOG.md` is showing incorrect year for releases since 2.10.6; giving the appearance of an inactive repository.


chalasr:
Good catch, thanks @tenfourzero.

PR #890, opened 2021-07-08
Removing @final for JWTUser constructor

abienvenu:
The "@final" annotation for the JWTUser constructor is somewhat inconsistent with the documentation which tells "Note: You can extend the default JWTUser class if that fits your needs." at https://github.com/lexik/LexikJWTAuthenticationBundle/blob/2.x/Resources/doc/8-jwt-user-provider.md


chalasr (approved):
Agree!

chalasr:
Thank you @abienvenu.

PR #894, opened 2021-07-14
Forward exception code to the response status code

theofidry:
It is not rare for some exceptions to have their status code taken as an HTTP status code. This avoids to have to override/decorate the authentication handler to mutate the returned response

### DIFF
--- a/Security/Http/Authentication/AuthenticationFailureHandler.php
+++ b/Security/Http/Authentication/AuthenticationFailureHandler.php
@@ -6,6 +6,7 @@ use Lexik\Bundle\JWTAuthenticationBundle\Event\AuthenticationFailureEvent;
 use Lexik\Bundle\JWTAuthenticationBundle\Events;
 use Lexik\Bundle\JWTAuthenticationBundle\Response\JWTAuthenticationFailureResponse;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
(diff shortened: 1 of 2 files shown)


Security/Http/Authentication/AuthenticationFailureHandler.php
```diff
@@ -33,13 +34,29 @@ public function __construct(EventDispatcherInterface $dispatcher)
...
 
         $this->dispatcher->dispatch($event, Events::AUTHENTICATION_FAILURE);
 
         return $event->getResponse();
     }
+
+    /**
+     * @param string|int $exceptionCode
+     */
+    private static function mapExceptionCodeToStatusCode($exceptionCode): int
```
chalasr, line 52:
```suggestion
    private static function mapExceptionCodeToStatusCode(int $exceptionCode): int
```
seems safe

chalasr (commented):
LGTM

Security/Http/Authentication/AuthenticationFailureHandler.php
theofidry: I'm mitigated: on one hand you are right it can only be an `int`, on the other `::getCode()` is coming from `Exception` which technically can be a string e.g. for `PDOException`...

chalasr: fair enough, let’s keep it as-is

chalasr (approved):

chalasr:
Thank you Théo.

PR #903, opened 2021-08-10
Implement `AuthenticatorInterface::createToken()` (Symfony 5.4)

chalasr:


Security/Authenticator/JWTAuthenticator.php
```diff
@@ -219,4 +220,17 @@ public function createAuthenticatedToken(PassportInterface $passport, string $fi
 
         return $token;
     }
+
+    public function createToken(Passport $passport, string $firewallName): TokenInterface
+    {
+        $token = parent::createToken($passport, $firewallName);
+
+        if (!$passport instanceof SelfValidatingPassport) {
```
mbabker, line 228: I don't think you'll need the `instanceof` check here.  The biggest need for it in `createAuthenticatedToken()` is with `PassportInterface` not having the attributes API, so someone using a custom passport with this authenticator could trigger a fatal error otherwise.

With that said, what you probably would want is to make sure `$passport->getAttribute('payload')` isn't null, as it seems like the important bit here is the data is attached to the passport and not necessarily the specific passport class.

PR #905, opened 2021-08-14
Changed `JWTAuthenticator::start` method return type to more generic `Response` type

### LINKED ISSUE #904: New JWTAuthenticator::start method requires hard return type of JWTAuthenticationFailureResponse
The new Symfony 5.3 `JWTAuthenticator::start` method has the return type of `JWTAuthenticationFailureResponse` when the event and all other methods require Symfony Response object. This breaks if your event subscribers set custom responses.

https://github.com/lexik/LexikJWTAuthenticationBundle/blob/06bd43f8aed787186e598aeb89be4c0dc2727650/Security/Authenticator/JWTAuthenticator.php#L75

aurimasniekis:
Not a big change just fixes an issue where the old code is upgraded to use the new JWTAuthenticator.

This fixes #904

### DIFF
--- a/Security/Authenticator/JWTAuthenticator.php
+++ b/Security/Authenticator/JWTAuthenticator.php
@@ -72,7 +72,7 @@ class JWTAuthenticator extends AbstractAuthenticator implements AuthenticationEn
     /**
      * {@inheritdoc}
      */
-    public function start(Request $request, AuthenticationException $authException = null): JWTAuthenticationFailureResponse
+    public function start(Request $request, AuthenticationException $authException = null): Response
     {
         $exception = new MissingTokenException('JWT Token not found', 0, $authException);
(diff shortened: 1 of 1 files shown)


chalasr (approved):

aurimasniekis:
I am not sure about`5.4@dev`build, I don't really see how is it related to my change.

chalasr:
Don’t worry about it, it’s not related

chalasr:
Thanks for fixing this bug @aurimasniekis.

fd6130:
Just encounter this issue on today. I think this patch should be release as soon as possible otherwise it break current project that have custom response. 

How do you think?  @chalasr

chalasr:
Thanks for the ping! I will draft the release tonight

PR #912, opened 2021-08-23
Added argument to AuthenticationSuccessHandler to stop token from being removed from response

### LINKED ISSUE #909: API and Cookie Authentication in the same Application
If one want to use the bundle in an application where user can authenticate per API (responding with a JSON body containing the token) and per form login (sending token as Cookie) the `Lexik\Bundle\JWTAuthenticationBundle\Security\Http\Authentication\AuthenticationSuccessHandler` cannot be used.

The reason is that the token is removed from the JSON response if there are any cookie providers configured (see [here](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/06bd43f8aed787186e598aeb89be4c0dc2727650/Security/Http/Authentication/AuthenticationSuccessHandler.php#L66)).

It would be nice if this could be changed.

stollr:
Fixes #909 (the handler cannot be used when API and form login are used in the same application).

To use this feature, define the handler as success handler on the firewall
```yaml
security:
  firewalls:
    main:
      json_login:
         check_path: my_login_check
         success_handler: lexik_jwt_authentication.handler.authentication_success
         failure_handler: lexik_jwt_authentication.handler.authentication_failure
```

And then define the fourth argument of the handler in your service.yaml:
```yaml
services:
    Lexik\Bundle\JWTAuthenticationBundle\Security\Http\Authentication\AuthenticationSuccessHandler:
        arguments:
            $removeTokenFromBodyWhenCookiesUsed: false
```

### DIFF
--- a/Security/Http/Authentication/AuthenticationSuccessHandler.php
+++ b/Security/Http/Authentication/AuthenticationSuccessHandler.php
@@ -27,15 +27,17 @@ class AuthenticationSuccessHandler implements AuthenticationSuccessHandlerInterf
 
     protected $jwtManager;
     protected $dispatcher;
+    protected $removeTokenFromBodyWhenCookiesUsed;
 
     /**
      * @param iterable|JWTCookieProvider[] $cookieProviders
      */
-    public function __construct(JWTTokenManagerInterface $jwtManager, EventDispatcherInterface $dispatcher, $cookieProviders = [])
+    public function __construct(JWTTokenManagerInterface $jwtManager, EventDispatcherInterface $dispatcher, $cookieProviders = [], bool $removeTokenFromBodyWhenCookiesUsed = true)
     {
         $this->jwtManager = $jwtManager;
         $this->dispatcher = $dispatcher;
         $this->cookieProviders = $cookieProviders;
+        $this->removeTokenFromBodyWhenCookiesUsed = $removeTokenFromBodyWhenCookiesUsed;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


chalasr (approved):

chalasr:
Thank you @naitsirch.

TjorvenB:
I was looking for a way to return the JWT token in both cookies and the response body. I spent an hour looking for this solution, and only found it because of the property name I found when I started looking at the code to see where I could extend it to allow this. So I thought I would also leave some searchable terms in this issue so other people can find it a bit quicker :-)
Thank you for the work!

chalasr:
I would welcome a PR to document this :)

TjorvenB:
I can't seem to get it to work with the service configuration as mentioned in the starting post. The container creates a second service definition for the AuthenticationSuccessHandler without the cookie providers. This service definition is not used because the bundle aliases the class. I tried aliasing the definition, hoping the cookie providers would be set later on. Quickly realised how idiotic that idea was.

I currently have it working by adding a new config value 'remove_token_from_body_when_cookies_used' which defaults to true. I have added the fourth argument to the service definition, defaulting to true. And a check in the LexikJWTAuthenticationExtension replacing the fourth parameter with false if the new config value is false.
If this approach is OK I can start a pull request for https://github.com/TjorvenB/LexikJWTAuthenticationBundle/tree/pr912 (with documentation)

I started from 2.x. Is this the correct branch to start from? There is also a 2.10.x branch.

fd6130:
> I can't seem to get it to work with the service configuration as mentioned in the starting post. The container creates a second service definition for the AuthenticationSuccessHandler without the cookie providers. This service definition is not used because the bundle aliases the class. I tried aliasing the definition, hoping the cookie providers would be set later on. Quickly realised how idiotic that idea was.
> 
> I currently have it working by adding a new config value 'remove_token_from_body_when_cookies_used' which defaults to true. I have added the fourth argument to the service definition, defaulting to true. And a check in the LexikJWTAuthenticationExtension replacing the fourth parameter with false if the new config value is false. If this approach is OK I can start a pull request for https://github.com/TjorvenB/LexikJWTAuthenticationBundle/tree/pr912 (with documentation)
> 
> I started from 2.x. Is this the correct branch to start from? There is also a 2.10.x branch.

I think 2.x is fine because this PR also add in 2.x

PR #914, opened 2021-08-26
Bundle breaks application if Symfony Console not installed

yivi:
Since the console command services are loaded only if `Symfony\Component\Console\Application` exists,  and later the extension attempts to modify the service defintion for `lexik_jwt_authentication.generate_keypair_command` (which would not be defined if Symfony Console is not installed), it can lead to breakage if Symfony Console is not installed.

Moved both loading the services configuration file and the definition change to the same conditional.


chalasr (approved):

chalasr:
Thank you @yivi.

PR #918, opened 2021-09-06
Only validate expiration date of incoming token if token_ttl is set

### LINKED ISSUE #889: Expiration date of incoming token is not validated if token_ttl is set to null
I understand that token_ttl should not be set to null, but it is possible and a non-expiring token is still a valid token.

Currently, the token_ttl setting is serving two purposes:
- it sets the expiry of newly created tokens
- it switches expiry validation on or off for all incoming tokens

I propose the token_ttl setting should be purely used to set the expiry of newly created tokens, and all incoming tokens should be validated according to the data in the token, regardless of the token_ttl setting.

Setting the ttl to null and validating a token with expiry date is not is not covered in the current tests as the tests do not use the 3rd parameter of the LoadedJWT. You will see that the test fails to reflect the token is expired if the 3rd parameter is set to false, which I think is incorrect. A token is expired based on the content of the token, not the system settings.

```php
public function testVerifiedWithExpiredPayload()
{
    $payload = $this->goodPayload;
    $payload['exp'] -= 3600;

    $jws = new LoadedJWS($payload, true);

    $this->assertFalse($jws->isVerified());
    $this->assertTrue($jws->isExpired());
}
```

To resolve this, a simple change is required to LoadedJWS, which ensures correct validation of all tokens

Current:
```php
public function __construct(array $payload, $isVerified, $hasLifetime = true, array $header = [], $clockSkew = 0)
    {
        $this->payload     = $payload;
        $this->header      = $header;
        $this->hasLifetime = $hasLifetime;
```

Proposed:
```php
public function __construct(array $payload, $isVerified, array $header = [], $clockSkew = 0)
    {
        $this->payload     = $payload;
        $this->header      = $header;
        $this->hasLifetime = isset($payload['exp']) ? true : false;
```

Would you like me to create a pull request for this?

pluk77:
fix #889


pluk77:
Is there a specific reason this patch has not been reviewed?

chalasr:
Only that my open-source time is limited. Thanks for the ping, this looks good to me, I need to check if it doesn't have "security" implications (e.g. that it does not open the door for tokens without `exp` while `token_ttl` is configured, etc.).
I'll be back on this as soon as I can.

pluk77:
@chalasr Thank you. I fully understand and appreciate your time constraints. If you do find security implications, please let me know what they are so we can find a better solution for the issue at hand.

chalasr:
Apologies again for the delay, this is a long standing topic.
Looking at this twice, I think this deserves a config option. It could be e.g. `verify_tokens_without_exp` with `false` as default. That setting would then be injected in the jws provider and passed to the `LoadedJWS` VO somehow to accept tokens with no `exp` claim when the option is set to true.
Would you mind updating your PR in that direction?
IMHO In production this only makes sense with a proper token invalidation mechanism, which I think we should add at some point (some storage abstraction alongside a revocation logic based on the `jti` claim).

pluk77:
I will have a look and amend my PR accordingly.

chalasr:
@pluk77 Do you want some help to fix the broken tests?

pluk77:
@chalasr Yes please!! That would be awesome.

pluk77:
Anybody able and willing to assist with getting these broken tests fixed so this PR can be merged?

chalasr:
Just pushed on your branch, PR is now green :) Can you please confirm that the new option still works as expected?

pluk77:
Thank you so much for your assistance here. I will go through the changes, but at first glance I think there are some changes which I do not understand the reasoning behind as they seem to contradict each other and some that will most likely still not allow for the different use-cases. I will run some tests and provide proper feedback and a potential solution for you to consider.

chalasr:
To make sure we are aligned, the goal of this PR is to allow validating tokens that have no `exp` claim when the option is set to `true`, right? Looking forward to your feedback.

pluk77:
I have made changes to your update to ensure the default settings are secure and that the ttl setting is not over-ruling the newly added validation setting. 

My reasoning is that the ttl in the settings is the default setting used to generate a token. But tokens can also be generated via commands or code and the EXP can be set via the payload (I have submitted a different PR to ensure that works correctly)

For that reason, the ttl setting should not influence if tokens without an exp should be validated or not as the validator does not know if those tokens were generated with custom or blank EXP. Ideally, the signature of a token should be used to determine / indicate if the EXP was removed from a token.

It works on my application and most test pass, but seeing that the default setting is now flagging tokens without EXP as invalid, there are a few test that fail.

I am not sure where in the test to add the new 'validate_tokens_without_ttl: false' setting as I am sure that will make these tests pass.

pluk77:
> To make sure we are aligned, the goal of this PR is to allow validating tokens that have no `exp` claim when the option is set to `true`, right? Looking forward to your feedback.

I understood it to be the other way around as this lines up with '$shouldCheckExpiration' in the LoadedJWS

No EXP + validate_tokens_without_ttl=false => results in token being validated without looking at EXP ($shouldCheckExpiration = false)
No EXP + validate_tokens_without_ttl=true => results in invalid token ($shouldCheckExpiration = true)
EXP + validate_tokens_without_ttl=false => results in EXP being validated ($shouldCheckExpiration = true)
EXP + validate_tokens_without_ttl=true=> results in EXP being validated ($shouldCheckExpiration = true)

chalasr:
Thanks for the explanation, I'll check this again asap this week.
I didn't look into your last changes yet, but please avoid changing existing tests to make sure they keep passing as they are. (The current behavior must not change to preserve backward compatibility, only new tests should be added to test the new behavior)

Services/JWSProvider/DefaultJWSProvider.php
```diff
@@ -45,16 +45,22 @@ class DefaultJWSProvider implements JWSProviderInterface
      * @var int
      */
     private $clockSkew;
+    
+    /**
+     * @var bool
+     */
+    private $validateTokensWithoutTtl;
```
chalasr: This class should not get new features, it is deprecated. Only LcobucciJWSProvider should benefit from this change

pluk77: > This class should not get new features, it is deprecated. Only LcobucciJWSProvider should benefit from this change

Understood. I have reverted all changes to DefaultJWSProvider. The additional parameter in the constructor of LcobucciJWSProvider should not cause an issue as the constructor is not defined in the providerInterface.

pluk77:
> I understood it to be the other way around as this lines up with '$shouldCheckExpiration' in the LoadedJWS

Perhaps we should change the setting name to something less confusing seeing that we both had a different understanding of the meaning? 

How about : validate_tokens_without_ttl_as_invalid = true (Default behavior, regardless of ttl value in yml file)

pluk77:
Please consider my new PR #1020

PR #923, opened 2021-09-16
Add 3 new getter method to JWTTokenAuthenticator

### LINKED ISSUE #782: Extending JWTTokenAuthenticator is not possible because variables are private
The documentation about [extending JWTTokenAuthenticator](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Resources/doc/6-extending-jwt-authenticator.md ) cannot be strictly followed because properties are stored as "private" and not "protected", and thus cannot be accessed in the extended class.
Source reference :
 https://github.com/lexik/LexikJWTAuthenticationBundle/blob/master/Security/Guard/JWTTokenAuthenticator.php
Redefining `getTokenExtractor()` for example is impossible, `$this->tokenExtractor` being private.
All properties should be defined as protected.

fd6130:
Fix #782 and #743

Add 3 new getter method for easy to reusing them in any child class. We don't have to explicit rewrite the same properties as parent class in custom guard authenticator anymore.

### DIFF
--- a/Security/Guard/JWTTokenAuthenticator.php
+++ b/Security/Guard/JWTTokenAuthenticator.php
@@ -250,6 +250,30 @@ class JWTTokenAuthenticator implements AuthenticatorInterface
     }
 
     /**
+     * @return JWTTokenManagerInterface
+     */
+    protected function getJwtManager()
+    {
+        return $this->jwtManager;
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


chalasr (approved):

chalasr:
Thank you @fd6130.

Issue #707, opened 2020-01-09
Configure multiple lexik_jwt_auth


Taluu:
Kinda linked to #504, but here goes ;

It could be great to be able to set multiple configurations for the config name, by using a prototype in the config. As does doctrine for example to be able to configure multiple entity managers / connections, so we can inject the jwt authentication we need.

e.g

```yaml
lexik_jwt_authentication:
	api:
	    secret_key: 'secret_key'
	    token_ttl: 3600
	    encoder:
	        signature_algorithm: HS256

	other:
	    secret_key: 'secret_key'
	    token_ttl: 60
	    encoder:
	        signature_algorithm: HS256
```

fl0cke:
Would be really nice to have this feature 👍

fd6130:
Yea this multiple config for jwt is nice because we can have multiple provider in security. Hope someone could implement this feature for this bundle :(

TonyBogdanov:
+1 for this!

In my setup I'm using JWT authentication for regular users, but I'm also using Google Cloud Pub/Sub with push endpoints, which I obviously need to firewall somehow. Google supports authenticating its requests with a JWT token generated for a service account, so all I need to do is verify that token against the service account.

The bundle works nicely to achieve this, the only problem is I can't use separate key pairs for Pub/Sub and my regular users, which is a disastrous security hole.

I'll have to explore @badrutdinovrr's custom authenticator workaround.

chalasr:
With the new Symfony authenticator-based security system, we are probably going to move most of the config to per-firewall configuration (with a `lexik_jwt: ~` configurable authenticator). Would that fit your use cases?

TonyBogdanov:
> With the new Symfony authenticator-based security system, we are probably going to move most of the config to per-firewall configuration (with a `lexik_jwt: ~` configurable authenticator). Would that fit your use cases?

Yes, that would actually be the best solution. I've been looking into the new approach and it really does simplify things a lot! Too bad it'll be mostly used in Symfony 6.0, which in my mind, is to be widely adopted in at least a couple of years...

Aweptimum:
> With the new Symfony authenticator-based security system, we are probably going to move most of the config to per-firewall configuration (with a `lexik_jwt: ~` configurable authenticator). Would that fit your use cases?

Yes +1 to this, just wanted to add my support. Thank you for this awesome bundle, though!

fd6130:
[...]

PR #928, opened 2021-09-28
Add support of multiple public keys to verify tokens with a set of keys

alexandre-daubois:
It may be convenient to use multiple public keys to verify token signature when using RS algorithm. It would allow to verify tokens coming from different providers.

Somehow related to #707


chalasr:
Thanks for the PR.
#707 is about allowing per-firewall configuration (including keys). Would that match your use case? Or do you need to support multiple keys for your entire application as this PR currently suggests?

alexandre-daubois:
Thanks for your reply! I don't think this would match our use case. Indeed, we would need to support multiple public keys for a single firewall. Tokens are coming from many providers, each with their own key, passing to a single firewall on a common endpoint.

HidouriAnis:
I currently have the same needs as you, what did you do as a solution?

alexandre-daubois:
@HidouriAnis for now, we did it like as it's in #707, in [this comment](https://github.com/lexik/LexikJWTAuthenticationBundle/issues/707#issuecomment-678762286). Even if this works, this is not super convenient to redefine all services like this. It's also very verbose.

chalasr (approved):

chalasr:
Thank you @alexandre-daubois.

PR #931, opened 2021-10-04
Fix: Only attempt split_cookie extraction if all of the cookies are present

### LINKED ISSUE #930: Invalidate split_cookies client-side
Like in #809, we use the split_cookie extractor to store the JWT token in two separate cookies, one of which is available in the JS frontend and one of which is not.
Also I read in https://github.com/lexik/LexikJWTAuthenticationBundle/issues/847#issuecomment-795776082 that on logout, JWT tokens should just be deleted client-side.

Now our frontend JS code can only delete the `jwt_hp` cookie, but not the `jwt_s` cookie, because that one is marked `httponly`. I see a few options how we could solve this:
* Ideally, the split_cookie extractor should be changed to require all the cookie parts. Currently it happily continues when some of the cookie parts are missing, but then joins all cookie values (whether present or not) with periods. This doesn't really make sense in any conceivable case: e.g. `jwt_hp=eyABCD.ey12341234` and a missing `jwt_s` cookie will result in the token `eyABCD.ey12341234.`, which can never be a valid JWT token because of the period in the end. For this reason, this wouldn't be a breaking change either, because no developers could have successfully used partial split cookies so far.
  I'd be happy to prepare a pull request to fix this if you're open to this change.
* Otherwise, I could listen to the JWT_INVALID event and delete all of my split cookies (by overwriting them with an expired expiry date).
* Or I could listen to some early Symfony event and delete all of my split cookies if not all of them are present.

I have found no easy way to extend / override the existing SplitCookieExtractor, otherwise that would be an option as well.

carlobeltrame:
Fixes #930 

As explained in #930, this isn't a breaking change, because setups with partial cookies couldn't have made sense for anyone before. Some examples:
* **2 cookies `jwt_hp` and `jwt_s`, first one missing**: Will result in `.eySignature`, which is not a valid JWT token (note the leading period)
* **2 cookies `jwt_hp` and `jwt_s`, second one missing**: Will result in `eyHeader.eyPayload.` which is not a valid JWT token (note the trailing period)
* **2 cookies `jwt_complete` and `optional_suffix`, second one missing**: Will result in `eyHeader.eyPayload.eySignature.` which is not a valid JWT token (note the trailing period)
* **3 cookies `jwt_h`, `jwt_p` and `jwt_s`, middle one missing**: Will result in `eyHeader..eySignature` which is not a valid JWT token (note the two consecutive periods)

So up until now, there is no way someone was successfully using the SplitCookieExtractor with only some of the cookies present.

### DIFF
--- a/Tests/TokenExtractor/SplitCookieTokenExtractorTest.php
+++ b/Tests/TokenExtractor/SplitCookieTokenExtractorTest.php
@@ -21,6 +21,10 @@ class SplitCookieTokenExtractorTest extends TestCase
         $this->assertFalse($extractor->extract($request));
 
         $request = new Request();
+        $request->cookies->add(['jwt_s' => 'testsignature']);
+        $this->assertFalse($extractor->extract($request));
+
+        $request = new Request();
         $request->cookies->add(['jwt_hp' => 'testheader.testpayload']);
         $request->cookies->add(['jwt_s' => 'testsignature']);
         $this->assertEquals('testheader.testpayload.testsignature', $extractor->extract($request));
--- a/TokenExtractor/SplitCookieExtractor.php
+++ b/TokenExtractor/SplitCookieExtractor.php
@@ -35,7 +35,7 @@ class SplitCookieExtractor implements TokenExtractorInterface
             $jwtCookies[] = $request->cookies->get($cookie, false);
         }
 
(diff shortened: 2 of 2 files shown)


chalasr (approved):

chalasr:
Good catch, thanks @carlobeltrame.

PR #932, opened 2021-10-04
Add return type in Configuration.php file to support symfony 6

eerison:
message error

```
Method "Symfony\Component\Config\Definition\ConfigurationInterface::getConfigTreeBuilder()" might add "TreeBuilder" as a native return type declaration in the future. Do the same in implementation "Lexik\Bundle\JWTAuthenticationBundle\DependencyInjection\Configuration" now to avoid errors or add an explicit @return annotation to suppress this message.
```


chalasr (approved):

chalasr:
Thank you @eerison.

PR #937, opened 2021-10-17
Symfony 6 Compatibility

mbabker:
This is the work needed for Symfony 6 compatibility.  The gist of the changes are:

- Allow Symfony 6 install
- Tweak the GitHub actions config, the tested versions are now set in a matrix so there's only one build config instead of duplicating it for each build
- Drops Symfony 5.3 support, the return signature change in `Symfony\Component\Security\Http\Authenticator\AuthenticatorInterface::authenticate()` between 5.3 and 5.4 is a bit tricky to deal with while keeping support for the non-LTS 5.3
- Updates `Lexik\Bundle\JWTAuthenticationBundle\DependencyInjection\Security\Factory\JWTAuthenticatorFactory` to address the security listener factory -> authenticator factory changes/deprecations in the SecurityBundle in 5.4
- Adds typehints or doc blocks as needed

### DIFF
--- a/DependencyInjection/Security/Factory/JWTAuthenticatorFactory.php
+++ b/DependencyInjection/Security/Factory/JWTAuthenticatorFactory.php
@@ -2,9 +2,7 @@
 
 namespace Lexik\Bundle\JWTAuthenticationBundle\DependencyInjection\Security\Factory;
 
-use Lexik\Bundle\JWTAuthenticationBundle\Security\Authenticator\JWTAuthenticator;
 use Symfony\Bundle\SecurityBundle\DependencyInjection\Security\Factory\AuthenticatorFactoryInterface;
-use Symfony\Bundle\SecurityBundle\DependencyInjection\Security\Factory\SecurityFactoryInterface;
 use Symfony\Component\Config\Definition\Builder\NodeDefinition;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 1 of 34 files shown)


DependencyInjection/Security/Factory/JWTAuthenticatorFactory.php
```diff
@@ -28,15 +26,15 @@ public function create(ContainerBuilder $container, $id, $config, $userProvider,
     /**
      * {@inheritdoc}
      */
-    public function getPosition()
+    public function getPriority(): int
```
mbabker, line 29: Technically, removing `getPosition()` could be a B/C break since the class isn't final or internal.  Same can be said for the return types below.

Security/Authentication/Token/JWTUserToken.php
```diff
@@ -6,12 +6,26 @@
 use Symfony\Component\Security\Core\User\UserInterface;
 use Symfony\Component\Security\Guard\Token\GuardTokenInterface;
 
+if (interface_exists(GuardTokenInterface::class)) {
```
mbabker, line 9: The compat layer here basically allows the user token class to be usable without the `security-guard` component.

chalasr:
🙏 Thanks a lot for working on this!

Given the diff on the public API is quite small, I think we can afford the added typehints without releasing a new major. 
About the `Response::setData(): static`, I fear we have to use the same `eval()` trick that you removed in `ForwardCompatTestCaseTrait`. Ok for you?

mbabker:
> I fear we have to use the same `eval()` trick that you removed in `ForwardCompatTestCaseTrait`. Ok for you?

Yeah, that would probably work out best.  I can mess with that later on.

DependencyInjection/Security/Factory/JWTAuthenticatorFactory.php
chalasr: I'm fine with this, the code is very recent and not likely to be extended. I'm gonna make it `@final` and mention the BC break in the changelog before tagging.

chalasr (approved):

chalasr:
Thanks a lot, @mbabker.

PR #951, opened 2021-11-12
Test  instanceof Passport instead of more restrictive SelfValidatingPassport

TristanPouliquen:
…assport

### DIFF
--- a/Security/Authenticator/JWTAuthenticator.php
+++ b/Security/Authenticator/JWTAuthenticator.php
@@ -240,8 +240,8 @@ class JWTAuthenticator extends AbstractAuthenticator implements AuthenticationEn
 
     public function createAuthenticatedToken(PassportInterface $passport, string $firewallName): TokenInterface
     {
-        if (!$passport instanceof SelfValidatingPassport) {
-            throw new \LogicException(sprintf('Expected "%s" but got "%s".', SelfValidatingPassport::class, get_debug_type($passport)));
+        if (!$passport instanceof Passport) {
+            throw new \LogicException(sprintf('Expected "%s" but got "%s".', Passport::class, get_debug_type($passport)));
         }
 
         $token = new JWTPostAuthenticationToken($passport->getUser(), $firewallName, $passport->getUser()->getRoles(), $passport->getAttribute('token'));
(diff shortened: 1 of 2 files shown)


Security/Authenticator/JWTAuthenticator.php
```diff
@@ -240,7 +240,7 @@ protected function loadUser(array $payload, string $identity): UserInterface
 
     public function createAuthenticatedToken(PassportInterface $passport, string $firewallName): TokenInterface
     {
-        if (!$passport instanceof SelfValidatingPassport) {
+        if (!$passport instanceof Passport) {
             throw new \LogicException(sprintf('Expected "%s" but got "%s".', SelfValidatingPassport::class, get_debug_type($passport)));
```
fd6130: You forgot to change this to `Passport::class`.

_Edit:_
Already updated before i write this.

fd6130 (commented):
Seems nice to me.

chalasr (approved):
Excellent 👍

Security/Authenticator/JWTAuthenticator.php
```diff
@@ -240,8 +240,8 @@ protected function loadUser(array $payload, string $identity): UserInterface
 
     public function createAuthenticatedToken(PassportInterface $passport, string $firewallName): TokenInterface
```
fd6130, line 241: >  (*PassportInterface are deprecated in Symfony 5.4)

What about this argument type here?

TristanPouliquen: I think @chalasr was talking about `UserPassportInterface` here, and not the generic `PassportInterface` (the * before makes me say this)

fd6130: Oh.. i didn't thought about it. But I think it is PassportInterface? You can check from here https://github.com/symfony/symfony/blob/5.4/CHANGELOG-5.4.md

Let's wait him for his opinion about this method argument type change.

TristanPouliquen: Of what I've seen the whole `createAuthenticatedToken` is going to be deprecated in favor of the simple `createToken` method in any case :)

chalasr: This method must remain the same for compatibility. Both `PassportInterface` and `createAuthenticatedToken()` are deprecated and removed in Symfony 6.0. 
Note that the reasoning for deprecating `PassportInterface` is that passports are not an extension point, one should use custom badges instead.

chalasr:
Thanks Tristan.

PR #961, opened 2021-12-01
Allow symfony/deprecations-contract v3.0

bravik:
Just to fix composer update issues, where composer prefers not to updated lexik bundle because it required deprecation-contracts v.2 and all other packages v3
image

```shell
dev@05584e3c9f0b:/app$ composer why-not lexik/jwt-authentication-bundle 2.14.1
lexik/jwt-authentication-bundle  v2.14.1     requires          symfony/deprecation-contracts (^2.4)
macroative/api                   dev-master  does not require  symfony/deprecation-contracts (but v3.0.0 is installed)
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
         "namshi/jose": "^7.2",
         "symfony/config": "^4.4|^5.3|^6.0",
         "symfony/dependency-injection": "^4.4|^5.3|^6.0",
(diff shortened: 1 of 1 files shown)


dougetovski:
I've encountered the same issue, thx @bravik

chalasr:
Thank you @bravik.

PR #966, opened 2021-12-06
fix getIterator compatible with php 8.1

eerison:


.github/workflows/ci.yml
```diff
@@ -44,6 +44,14 @@ jobs:
...
                       composer-flags: ''
                       can-fail: false
+                    - php: 8.1
+                      symfony: 5.4.*@dev
+                      composer-flags: ''
+                      can-fail: false
+                    - php: 8.1
+                      symfony: 6.0.*@dev
+                      composer-flags: ''
+                      can-fail: false
```
chalasr: I'm reverting this changes as the CI is updated in #968.

composer.json
```diff
@@ -37,7 +37,7 @@
         }
     ],
     "require": {
-        "php": ">=7.1",
+        "php": "^7.1 || ^8.0",
```
chalasr: I'm going to revert this as I prefer keeping it open.

phpunit.xml.dist
```diff
@@ -28,6 +28,6 @@
     </filter>
 
     <php>
-        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=0&amp;max[direct]=27"></env>
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=17&amp;max[direct]=27"></env>
```
chalasr: I'm reverting this one as it should not be needed (deprecations should be fixed instead).

chalasr:
Thanks for fixing this bug @eerison.

PR #973, opened 2021-12-29
Translate message errors

flohw:
Hi @chalasr,

The exception messages are not translated even if the locale was set.
I largely ~copied~ inspired my code from this PR: symfony/symfony#38037 to translate the message.

Let me know if any changes need to be done.

### DIFF
--- a/Security/Http/Authentication/AuthenticationFailureHandler.php
+++ b/Security/Http/Authentication/AuthenticationFailureHandler.php
@@ -10,6 +10,7 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Security\Core\Exception\AuthenticationException;
 use Symfony\Component\Security\Http\Authentication\AuthenticationFailureHandlerInterface;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
+use Symfony\Contracts\Translation\TranslatorInterface;
 
 /**
  * AuthenticationFailureHandler.
@@ -23,9 +24,15 @@ class AuthenticationFailureHandler implements AuthenticationFailureHandlerInterf
      */
     protected $dispatcher;
 
-    public function __construct(EventDispatcherInterface $dispatcher)
+    /**
+     * @var TranslatorInterface|null
+     */
+    private $translator;
+
+    public function __construct(EventDispatcherInterface $dispatcher, TranslatorInterface $translator = null)
     {
         $this->dispatcher = $dispatcher;
+        $this->translator = $translator;
     }
 
     /**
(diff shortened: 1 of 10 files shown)


flohw:
I just noted that on the original PR translations are made in the authenticator. It may be a better choice as the listener can be changed by the developer.

Let me know if I missed something and should update to translate in the authenticators too or only in them.

Thank you

[EDIT] also, not sure why the tests fail...

Security/Http/Authentication/AuthenticationFailureHandler.php
```diff
@@ -23,18 +24,31 @@ class AuthenticationFailureHandler implements AuthenticationFailureHandlerInterf
      */
     protected $dispatcher;
 
+    /**
+     * @var TranslatorInterface|null
+     */
+    protected $translator;
+
     public function __construct(EventDispatcherInterface $dispatcher)
```
chalasr: Let's use a nullable constructor parameter instead of the setter

Security/Http/Authentication/AuthenticationFailureHandler.php
```diff
@@ -23,18 +24,31 @@ class AuthenticationFailureHandler implements AuthenticationFailureHandlerInterf
      */
     protected $dispatcher;
 
+    /**
+     * @var TranslatorInterface|null
+     */
+    protected $translator;
```
chalasr: please make it `private` (yes that's inconsistent but we don't want to provide such extension points anymore)

chalasr (changes requested):
Thanks a lot for the PR. Here are some suggestions :)

chalasr:
Don't worry about the failing tests, it's unrelated. I need to look into them, sorry about that.

flohw:
Thank you for this quick review! I updated the code accordingly. (sorry for the multiple push force, I struggled with my gpg key...)

Also note about this comment above:
> I just noted that on the original PR translations are made in the authenticator. It may be a better choice as the listener can be changed by the developer.
> 
> Let me know if I missed something and should update to translate in the authenticators too or only in them.

Should I update the other place where `strtr` is used?

chalasr (approved):

chalasr:
Sorry I was unsure to understand your question, got it now.
If you could do the same in the jwt authenticator that would be awesome (both are good, so what you already did is valid!)

flohw:
Hi @chalasr,

Thanks I just added translator in both authenticators (JWT and JWTToken).

chalasr (approved):

flohw:
Hi @chalasr 
I just saw the new release and rebase my branch.
The failing tests have changed. I am not sure if it's on my side to repair especially for the PHP8/SF5.3 couple.

Thanks

chalasr:
No worries, it's due to a Symfony bug revealed by your changes. See https://github.com/symfony/symfony/pull/44918 for the fix.

chalasr:
Thanks @flohw for working on this feature, this is much appreciated.

PR #976, opened 2022-01-09
Fix authentication with integer as useridentifier

Floruzus:
Fix authentication with integer as useridentifier caused argument type error with UserBadge (Maybe only with Symfony >= 6.0)


fd6130:
Good catch, mostly it will affect Symfony >= 5.2 with new authenticator system enabled.

chalasr:
Thank you @Floruzus.

PR #977, opened 2022-01-12
Fix AuthenticationFailureListener example

Seb33300:


chalasr (approved):

chalasr:
Good catch, thanks @Seb33300.

PR #982, opened 2022-01-24
Fix a type related depreciation with php 8.1

fsalmen-nx:
With php 8.1 passing null as a second parameter to the explode function is deprecated. 
See https://www.php.net/manual/en/migration81.deprecated.php
`$request->headers->get($this->name);` can return null so i suggest a simple cast to fix this deprecation.


chalasr (approved):
LGTM. Thanks for the contrib!

chalasr:
Thank you @RiffFred.

PR #999, opened 2022-04-01
Unify audience claim

aerrasti:
The 'aud' part of the payload is being converted to array, thats why we need to use UnifyAudience claim formatter to create lcobucci token.

Before the change:
```json
{
  "iss": "iss",
  "aud": [
     "audience"
  ],
  "sub": "sub",
}
```
After the change
```json
{
  "iss": "iss",
  "aud": "audience",
  "sub": "sub",
}
```
Check out the `default` method:

https://github.com/lcobucci/jwt/blob/4.2.x/src/Encoding/ChainedFormatter.php

:)

### DIFF
--- a/Services/JWSProvider/LcobucciJWSProvider.php
+++ b/Services/JWSProvider/LcobucciJWSProvider.php
@@ -4,8 +4,8 @@ namespace Lexik\Bundle\JWTAuthenticationBundle\Services\JWSProvider;
 
 use Lcobucci\Clock\SystemClock;
 use Lcobucci\JWT\Builder;
+use Lcobucci\JWT\Encoding\ChainedFormatter;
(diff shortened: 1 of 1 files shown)


jordisala1991:
This change is needed to be compatible with software that only accepts audiences generated as strings like https://github.com/jitsi

chalasr (approved):
👍

chalasr:
Thank you @aerrasti.

jordisala1991:
Can you do a patch release including this one? Many thanks 🙏 !

chalasr:
Released in [v2.15.1](https://github.com/lexik/LexikJWTAuthenticationBundle/releases/tag/v2.15.1)!

Issue #676, opened 2019-08-05
Make "token_ttl" a parameter during "create" new token


toxaco:
I wonder if there is a way to change the token_ttl in a "per token" bases? 

For example, once you call "$jwt->create($user, $token_ttl)" to get a token based on the provided "token_ttl" and not in the global config (set in parameters.yml or .env).

If you provide your preferences/restrictions I can submit a request with these changes.

franzwilding:
I'm also looking for exactly this feature. I want to create short and long loving tokens.

franzwilding:
@toxaco I just figured out, that you can set the exp in an JWT_CREATED event subscriber: 

```
$data['exp'] = time() + 12334;
```

This is not exactly what I need, but with a little ugly hack you can use the user object to pass context information to the event and set the exp based on that information.

pluk77:
You might be able to do this by setting the default ttl to null and passing the ttl in the payload when creating the token.

```yaml
lexik_jwt_authentication:
    token_ttl:   null
```

```php
$payload['exp'] = time() + 36000;
$JWTManager->createFromPayload($user,$payload);
```

This is possible because the create method of DefaultJWSProvider and LcobucciJWSProvider both only use ```$this->ttl``` if ```$payload['exp']``` is not set.

pluk77:
I found an issue with my previous solution as setting the token_ttl to null will bypass the exp validation of the token completely. see #889

jawira:
I'm also looking for this feature.
Related https://github.com/lexik/LexikJWTAuthenticationBundle/issues/682

pluk77:
@jawira Perhaps you can assist with having a look at the [PR918](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/918). I can not seem to get those tests working, which prevents the PR from being actioned.

PR #1007, opened 2022-04-20
Allow for creation of tokens without exp

pluk77:
Allow exp=0 to be set in the payload during creation, which will create a token without an EXP. Related to #676 and #889.

### DIFF
--- a/Command/GenerateTokenCommand.php
+++ b/Command/GenerateTokenCommand.php
@@ -45,8 +45,9 @@ class GenerateTokenCommand extends Command
         $this
             ->setName(static::$defaultName)
             ->setDescription('Generates a JWT token')
-            ->addArgument('username', InputArgument::REQUIRED)
-            ->addOption('user-class', 'c', InputOption::VALUE_REQUIRED)
+            ->addArgument('username', InputArgument::REQUIRED, 'Username of user to be retreived from user provider')
+            ->addOption('ttl', 't', InputOption::VALUE_REQUIRED, 'Ttl in seconds to be added to current time. If not provided, the ttl configured in the bundle will be used. Use 0 to generate token without exp')
+            ->addOption('user-class', 'c', InputOption::VALUE_REQUIRED, 'Userclass is used to determine which user provider to use')
         ;
     }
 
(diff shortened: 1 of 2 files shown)


pluk77:
Hello @chalasr . Any idea yet when you will have some time to review and consider this PR?

chalasr:
Can we use`--ttl=0` to skip adding the exp claim instead? And remove the `payload` option? I'm not sure we need it. At least it deserves its own discussion.

pluk77:
I can sure try and make it work with `--ttl=0`. As for the payload, to us it is important as we use the payload to encapsulate additional meta-data. Our system can not do without, but I can remove it and make my own command that includes this specific feature. I will separate the two features into separate PRs so you can decide.

pluk77:
I have removed the payload option from this patch. I still believe it is a good addition. We use a generated JWT token as an access token between two applications that reside on the same infrastructure. The token never leaves our network nor does it go to the user. But because we need a bit more context in the token, we add that via the payload. When we generate the token programmatically, we can add the payload without problems. Now I need a solution for the payload to be added by tokens generated from the command-line. 

I have created a second PR, but now with the additional payload option for your consideration. If you choose not to allow the additional payload, I can always create my own command to implement this, but I would really like this PR to be considered for next release.

Command/GenerateTokenCommand.php
```diff
@@ -44,9 +44,10 @@ protected function configure()
     {
         $this
             ->setName(static::$defaultName)
-            ->setDescription('Generates a JWT token')
-            ->addArgument('username', InputArgument::REQUIRED)
-            ->addOption('user-class', 'c', InputOption::VALUE_REQUIRED)
+            ->setDescription('Generates a JWT token with optional payload')
+            ->addArgument('username', InputArgument::REQUIRED, 'Username of user to be retreived from user provider')
+            ->addArgument('ttl', InputArgument::OPTIONAL, 'Ttl in seconds to be added to current time. If not provided, the ttl configured in the bundle will be used. Use 0 to generate token without exp', null)
```
chalasr: What about making it an option instead?

pluk77: I have no objection to that. Will amend accordingly.

chalasr:
Thank you @pluk77.

PR #1008, opened 2022-04-20
Allow additional public keys to be set by value, as well as by path

MatthewHepburn:
https://github.com/lexik/LexikJWTAuthenticationBundle/pull/928 added an excellent feature, allowing additional public keys to be specified.

However, unlike the main public key, this can only be specified as a file path, rather than being passed directly as a string. The ability to pass keys as strings is useful when setting keys via environment variables, so I want to be able to configure the additional public keys as either file paths or as raw strings.

### DIFF
--- a/Services/KeyLoader/AbstractKeyLoader.php
+++ b/Services/KeyLoader/AbstractKeyLoader.php
@@ -48,7 +48,7 @@ abstract class AbstractKeyLoader implements KeyLoaderInterface
 
         foreach ($this->additionalPublicKeys as $key) {
             if (!$key) {
-                throw new \RuntimeException(sprintf('Additional public key is not set correctly. Check the "lexik_jwt_authentication.additional_public_keys" configuration key', $key));
+                continue;
             }
             if (is_file($key) && !is_readable($key)) {
                 throw new \RuntimeException(sprintf('Additional public key "%s" is not readable. Did you correctly set the "lexik_jwt_authentication.additional_public_keys" configuration key?', $key));
--- a/Services/KeyLoader/AbstractKeyLoader.php
+++ b/Services/KeyLoader/AbstractKeyLoader.php
@@ -47,8 +47,11 @@ abstract class AbstractKeyLoader implements KeyLoaderInterface
         $contents = [];
 
         foreach ($this->additionalPublicKeys as $key) {
-            if (!$key && (!is_file($key) || !is_readable($key))) {
-                throw new \RuntimeException(sprintf('Additional public key "%s" does not exist or is not readable. Did you correctly set the "lexik_jwt_authentication.additional_public_keys" configuration key?', $key));
+            if (!$key) {
+                throw new \RuntimeException(sprintf('Additional public key is not set correctly. Check the "lexik_jwt_authentication.additional_public_keys" configuration key', $key));
+            }
+            if (is_file($key) && !is_readable($key)) {
+                throw new \RuntimeException(sprintf('Additional public key "%s" is not readable. Did you correctly set the "lexik_jwt_authentication.additional_public_keys" configuration key?', $key));
             }
 
             $contents[] = is_file($key) ? file_get_contents($key) : $key;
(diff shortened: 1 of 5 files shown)


chalasr:
Sounds good! Let me know when you think this is ready

Services/KeyLoader/AbstractKeyLoader.php
```diff
@@ -47,8 +47,11 @@ public function getAdditionalPublicKeys(): array
         $contents = [];
 
         foreach ($this->additionalPublicKeys as $key) {
-            if (!$key || !is_file($key) || !is_readable($key)) {
-                throw new \RuntimeException(sprintf('Additional public key "%s" does not exist or is not readable. Did you correctly set the "lexik_jwt_authentication.additional_public_keys" configuration key?', $key));
+            if (!$key) {
+                continue;
```
chalasr, line 51: what's the motivation for this change?

MatthewHepburn: I want to support having a dynamic number of additional keys using a static configuration. For example 1 additional key while a key rotation is in progress, and 0 additional keys otherwise.

I haven't been able to come up with a better way of expressing that than
```yaml
lexik_jwt_authentication:
  public_key: '%env(MAIN_KEY)%'
  additional_public_keys:
    - '%env(string:default::SECONDARY_KEY)%'
```

But then we end up with this invalid array element when `SECONDARY_KEY` is not set that needs to be skipped for that to work.

I did consider trying to set the whole array via JSON like
```yaml
lexik_jwt_authentication:
  public_key: '%env(MAIN_KEY)%'
  additional_public_keys: '%env(json:SECONDARY_KEY_JSON)%'
```

But that's not allowed - we get the error `A dynamic value is not compatible with a "Symfony\Component\Config\Definition\PrototypedArrayNode" node type at path "lexik_jwt_authentication.additional_public_keys"`.

I have limited experience with the Symfony config system, so perhaps I'm missing a better way of expressing this. I would be happier keeping the behaviour of throwing on encountering a falsy value and handling this situation at the config level instead, but I'm not seeing a way to do that.

chalasr: Let me think about it :)

Ardenexal:
@chalasr @MatthewHepburn any update on this I running into the same issue where I want to pass through the Public key string rather than the file location and this PR would fix that issue

Services/KeyLoader/AbstractKeyLoader.php
1ed: We also have a similar use case.

1ed:
What is missing to get this merged?

TekSign:
Is there any update on this change please ? Because this would be very helpful and all seems ready.

Services/KeyLoader/AbstractKeyLoader.php
danieleartico: Hello! I'm facing a similar problem: using additional string certs instead of files to verify tokens.

IMHO, looking at the code, the original intention of the `AbstractKeyLoader` was to allow both cases because the `$key` is expected to be stored in the `$contents` array in its raw format even if the file is missing. Furthermore a similar behavior is defined in the `getSigningKey()` method.

Unfortunately, due to the (wrong?) `if` statement at the beginning of the loop, this situation never happens because when the key is properly set but it does not point to a file, then the exception is thrown.

@MatthewHepburn beware that what you tried to achieve in your first [commit](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1008/commits/86c484f46c9b998492cda2c24ecbde271bf853eb) did not solve this problem because, when the `$key` is properly configured, `!$key` always solves to false and the condition on the right side is never considered 🙂

Also, I'd leave the `\RuntimeException` if the string is empty, because in my opinion it is a config-related issue and it should not be addressed here 🤔

That said, I'd change the code as following:
```php
if (!$key || (is_file($key) && !is_readable($key)) {
    throw new \RuntimeException(...);
}
```

What do you think?

PR #1011, opened 2022-04-23
[Docs] Tweaks and improvements

javiereguiluz:
This PR includes the following:

* Some minor formatting and typo fixes
* Change the `<h3>` underline to `~~~` instead of `^^^` to comply with Symfony Docs practices
* Change the `<h4>` underline to `...` instead of `'''` to comply with Symfony Docs practices
* Reindent all code blocks (they used 3 white spaces but we need 4)
* Use `code-block:: terminal` instead of `code-block:: bash` when we want to display some console command
* Use `.. note::` and `.. caution::` admonitions
* Fix the syntax of some internal references
* Display the "Before Symfony 5.3" and "When using Symfony 5.3 or higher" texts as headings to not confuse readers

### DIFF
--- a/Resources/doc/1-configuration-reference.rst
+++ b/Resources/doc/1-configuration-reference.rst
@@ -8,88 +8,88 @@ Minimal configuration
 ~~~~~~~~~~~~~~~~~~~~~
 
 Using RSA/ECDSA
-^^^^^^^^^^^^^^^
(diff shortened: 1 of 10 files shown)


chalasr (approved):

chalasr:
Thanks a ton, Javier!

PR #1015, opened 2022-04-26
Fix ECDSA algo names in LcobucciJWSProvider

ControlAltBen:
Hi,

While trying to use ECDSA signature, I encountered `The algorithm "ES256" is not supported by Lexik\Bundle\JWTAuthenticationBundle\Services\JWSProvider\LcobucciJWSProvider`.

This name mismatch is pretty clear observing the `GenerateKeyPairCommand` command, which defines `ES256`, `ES384` and `ES512`  in its `ACCEPTED_ALGORITHMS` constant.

This PR fixes this issue and allows usage of ESxxx algorithms.


chalasr:
Nice catch!

chalasr:
Thank you @ControlAltBen.

PR #1016, opened 2022-04-28
Update token claims list on doc

romlly:
The JWT default claims seems out of date so I updated them.


chalasr (approved):

chalasr:
Good catch, thanks @romlly.

Issue #944, opened 2021-11-02
Psalm issue after update to `2.14.1` 


tarlepp:
Noticed that Psalm complains following;
```
psalm: MissingDependency: Lexik\Bundle\JWTAuthenticationBundle\Response\JWTAuthenticationFailureResponse depends on class or interface lexik\bundle\jwtauthenticationbundle\response\jwtcompatauthenticationfailureresponse that does not exist
```

Related code;
```php
public function onAuthenticationFailure(Request $request, AuthenticationException $exception): Response
{
    $event = new AuthenticationFailureEvent(
        $exception,
        new JWTAuthenticationFailureResponse(
            $this->translator->trans('Invalid credentials.', [], 'security')
        )
    );

    $this->dispatcher->dispatch($event);

    return $event->getResponse();
}
```

Although everything works as expected - Should I raise an issue to Psalm repository about this?

chalasr:
Thanks for the report. This is due to a compatibility layer implying some ugly hacks, I don't think psalm should care.
I'll try to fix that here.

tarlepp:
@chalasr awesome! And really big thanks for quick replies on issues and PR's to fix those!

tarlepp:
And just noticed that after updating PHPStan to version 1.6.0 it also complains at this.

```
phpstan: Parameter #2 $response of class Lexik\Bundle\JWTAuthenticationBundle\Event\AuthenticationFailureEvent constructor expects Symfony\Component\HttpFoundation\Response, Lexik\Bundle\JWTAuthenticationBundle\Response\JWTAuthenticationFailureResponse given.
```

So currently I'm using following to "fix" this:

```php
public function onAuthenticationFailure(Request $request, AuthenticationException $exception): Response
{
    /**
     * @psalm-suppress MissingDependency, InvalidArgument
     */
    $event = new AuthenticationFailureEvent(
        $exception,
        new JWTAuthenticationFailureResponse( // @phpstan-ignore-line
            $this->translator->trans('Invalid credentials.', [], 'security')
        )
    );

    $this->dispatcher->dispatch($event);

    return $event->getResponse();
}
```

chalasr:
I wouldn't have expected that, but our hack-ish BC layers may be handled by PHPStan https://github.com/phpstan/phpstan/issues/7157. Let's see how it goes and eventually propose the same for psalm

ondrejmirtes:
[...]

PR #1019, opened 2022-05-03
Fix lexik#944: Separate CompatFailureResponse from FailureResponse

GErpeldinger:
Hello,

Following ondrejmirtes' feedback on issue #944, here is a little fix that might do the trick.

It seems that just separating the classes is not enough for psalm, and that you have to blow up the eval(), which doesn't seem to me to bring any particular difficulty?

update: Oh my bad, it makes the tests fail in 7.1 & 7.4


chalasr (approved):
What a great contribution. Thank you very much for tracking down this bug and fixing it.

GErpeldinger:
I changed the `&&` to `AND` until the problem is solved on psalm side.

chalasr:
Thank you @GErpeldinger.

PR #1024, opened 2022-05-16
[Doc] Replace tilde in paths with carets

Sajito:
The paths started with a tilde, which doesn't seem to make sense. I think they were changed to tilde by mistake, so I've replaced them by carets again.


chalasr (approved):

chalasr:
Good catch, thanks @Sajito.

PR #1035, opened 2022-06-08
Code migration helper

Spomky:
This PR introduces Rector in order to help migrating and upgrading code using a few set of rules e.g. for PHP or Symfony.

### DIFF
--- a/Command/GenerateTokenCommand.php
+++ b/Command/GenerateTokenCommand.php
@@ -55,7 +55,7 @@ class GenerateTokenCommand extends Command
      *
      * @return int
      */
-    protected function execute(InputInterface $input, OutputInterface $output)
+    protected function execute(InputInterface $input, OutputInterface $output): int
     {
         if ($this->userProviders instanceof \Countable && 0 === \count($this->userProviders)) {
             throw new \RuntimeException('You must have at least 1 configured user provider to generate a token.');
--- a/Tests/Functional/app/AppKernel.php
+++ b/Tests/Functional/app/AppKernel.php
@@ -2,7 +2,11 @@
 
 namespace Lexik\Bundle\JWTAuthenticationBundle\Tests\Functional;
 
+use Lexik\Bundle\JWTAuthenticationBundle\LexikJWTAuthenticationBundle;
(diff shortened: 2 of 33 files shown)


Command/GenerateTokenCommand.php
```diff
@@ -2,6 +2,9 @@
 
 namespace Lexik\Bundle\JWTAuthenticationBundle\Command;
 
+use Traversable;
+use Countable;
+use RuntimeException;
```
chalasr: same CS as Symfony here :) I don't like having use statements for things coming from the global namespace

Spomky: I understand. I removed the rule and reverted the changes

Tests/Functional/app/AppKernel.php
```diff
@@ -37,12 +41,12 @@ public function __construct($environment, $debug, $testCase = null)
...
-            new \Symfony\Bundle\SecurityBundle\SecurityBundle(),
-            new \Lexik\Bundle\JWTAuthenticationBundle\LexikJWTAuthenticationBundle(),
-            new \Lexik\Bundle\JWTAuthenticationBundle\Tests\Functional\Bundle\Bundle(),
-        ];
+        return array(
+            new FrameworkBundle(),
+            new SecurityBundle(),
+            new LexikJWTAuthenticationBundle(),
+            new Bundle(),
+        );
```
chalasr:
```suggestion
        return [
            new FrameworkBundle(),
            new SecurityBundle(),
            new LexikJWTAuthenticationBundle(),
            new Bundle(),
        ];
```

chalasr (approved):
Cool! One minor comment

Tests/Functional/app/AppKernel.php
Spomky: Strange... Rector reverted to to the old style `array(` for no reason.
It is now fixed.

Spomky:
In the last commit, I took the opportunity to change the `$this->assertSame(xxx, $response->getStatusCode());` into `$this->assertResponseStatusCodeSame(xxx);`

chalasr:
Thank you @Spomky.

PR #1038, opened 2022-06-13
PHP_CS removed in favour of ECS

Spomky:
This PR removes HP_CodeSniffer in favor of ECS (symplify/easy-coding-standard)

ECS includes PHP_CS and PHP CS Fixer rulesets.
The proposed configuration is the same as the one that was used, except that it relies on PSR12 instead of the deperacted PSR1 and PSR2 (see this [migration guide](https://tomasvotruba.com/blog/2018/06/04/how-to-migrate-from-php-code-sniffer-to-easy-coding-standard/#comment-4086561141)).

In addition, a Github Action workflow is added to run both Rector and ECS and detect code lines that could be improved.


chalasr:
Thank you @Spomky.

PR #1041, opened 2022-06-26
WebToken support integration

Spomky:
This PR aims to integrate Web-Token for issuing and verifying access tokens.
The two processes are separated so that it is possible to issue a token on one server and verify them on another.

The configuration looks like as follow

```yaml
lexik_jwt_authentication:
    encoder:
        service: lexik_jwt_authentication.encoder.web_token #New encoder
    access_token_issuance:
        enabled: true
        signature: # Access tokens are always signed
            algorithm: 'HS256' # Signature/MAC algorithm.
            key: '%env(LEXIK_SIGNATURE_KEY)%' # Signature/MAC key (private or shared key, JWK format)
        encryption: # Access token may be encrypted
            enabled: true
            key_encryption_algorithm: 'A256GCMKW' # Key Encryption Algorithm.
            content_encryption_algorithm: 'A256GCM' # Content Encryption Algorithm.
            key: '%env(LEXIK_ENCRYPTION_KEY)%' # Encryption key (public or shared key, JWK format)
    access_token_verification:
        enabled: true
        signature: # Received Access tokens shall always be signed
            allowed_algorithms: ['HS256', 'RS256'] # List of allowed signature/MAC algorithms
            keyset: '%env(LEXIK_ALLOWED_SIGNATURE_KEYSET)%' # Signature verification keys (public or shared keys, JWKSet format)
        encryption:
            enabled: true: # Received Access tokens may be encrypted
            allowed_key_encryption_algorithms: ['A256GCMKW', 'ECDH-ES+A128KW'] # List of allowed key encryption algorithms
            allowed_content_encryption_algorithms: ['A256GCM', 'A128CBC-HS256'] # List of allowed content encryption algorithms
            keyset: '%env(LEXIK_ALLOWED_ENCRYPTION_KEYSET)%' # Decryption keys (private or shared keys, JWKSet format)
```

> **Note**
> To work with this feature, the application shall install the associated bundle and the algorithms to be used:

```shell
composer require web-token/jwt-bundle
composer require web-token/jwt-checker

composer require web-token/jwt-signature-algorithm-hmac
composer require web-token/jwt-encryption-algorithm-aesgcm
composer require web-token/jwt-encryption-algorithm-aesgcmkw
```

Complete lists of supported algorithms:
* Signature/MAC: https://web-token.spomky-labs.com/the-components/signed-tokens-jws/signature-algorithms
* Key encryption: https://web-token.spomky-labs.com/the-components/encrypted-tokens-jwe/encryption-algorithms#key-encryption
* Content encryption: https://web-token.spomky-labs.com/the-components/encrypted-tokens-jwe/encryption-algorithms#content-encryption

> **Warning**
> This feature is **NOT** compatible with the cookie split feature!

* [x] Documentation
* [x] JWT Builder
    * [x] Signature alg+key
    * [x] Encryption algs+key
    * [x] Additional claims and header
    * ~Allow additional claims and header to be disable~ (removed as too risky)
    * [x] Allow custom JWE header
* [x] JWT Loader
    * [x] Decryption algs+keyset
    * [x] Verification algs+keyset
    * [x] JWS claims and header checkers
    * [x] JWE header checkers
    * [x] Allow non-encrypted tokens even if encryption support is enable
* [x] Commands
    * [x] Configuration migration + Key conversion
    * [x] Encryption Support Helper
    * ~Keyset Rotation Helper~ (will be part of another PR)
* [x] Tests
    * [x] Success
        * [x] Token issuance
        * [x] Token verification
    * [x] Failure
        * [x] Not encrypted
        * [x] Cannot be decrypted
            * [x] Unsupported key encryption algorithm
            * [x] Unsupported content encryption algorithm
            * [x] Missing decryption key
        * [x] Bad content
        * [x] Bad signature
        * [x] Unsupported signature algorithm
        * [x] Missing verification key
        * [x] Time sensitive (expired, not yet...)
        * [x] Mandatory claims are missing


Spomky:
Hi @chalasr and all,

It looks good to me. Can you please take time to review this PR and let me know if you see issues, feature or anything that should be addressed here.

Many thanks,
Regards.

chalasr:
Thanks for the amazing work @Spomky. Review on its way :)

Tests/Functional/Command/MigrateConfigCommandTest.php
```diff
@@ -0,0 +1,39 @@
...
+
+use Lexik\Bundle\JWTAuthenticationBundle\Tests\Functional\TestCase;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Tester\CommandTester;
+
+/**
+ * CheckOpenSSLCommandTest.
+ *
+ * @author Nicolas Cabot <n.cabot@lexik.fr>
```
chalasr:
```suggestion
```

Services/WebToken/AccessTokenLoader.php
```diff
@@ -0,0 +1,166 @@
...
+        if (!is_array($data)) {
+            throw new JWTDecodeFailureException(JWTDecodeFailureException::INVALID_TOKEN, 'Invalid payload. The token shall contain claims.');
+        }
+
+        return $data;
+    }
+
+    private function loadJWE(string $token): string
+    {
+        if ($this->jweLoader === null) {
```
chalasr:
```suggestion
        if (!$this->jweLoader) {
```
per our cs :) please do the same for all similar checks added here if possible

Services/WebToken/AccessTokenLoader.php
```diff
@@ -0,0 +1,166 @@
...
+     * @param string[] $claimChecker
+     * @param string[] $jwsHeaderChecker
+     * @param string[] $mandatoryClaims
+     * @param string[] $signatureAlgorithms
+     * @param string $signatureKeyset
+     * @param string[] $jweHeaderChecker
+     * @param string[] $keyEncryptionAlgorithms
+     * @param string[] $contentEncryptionAlgorithms
+     * @param string|null $encryptionKeyset
+     */
```
chalasr: phpdoc params that tell the exact same thing as the type declaration can be removed

Services/WebToken/AccessTokenLoader.php
```diff
@@ -0,0 +1,166 @@
...
+    private $encryptionKeyset;
+
+    /**
+     * @var string[]
+     */
+    private $mandatoryClaims;
+
+    /**
+     * @var bool|null
+     */
```
chalasr: let's rely on IDEs inferring the type from constructor params and remove property types (sorry if you did this for consistency, the whole codebase will be refactored at some point)

Resources/doc/index.rst
```diff
@@ -35,6 +35,15 @@ Register bundle into ``config/bundles.php`` (Flex did it automatically):
 Generate the SSL keys
 ~~~~~~~~~~~~~~~~~~~~~
 
+.. note::
+
+    For the version 2.xx of this bundle, you can use Web-Token and generate
```
chalasr: A link to the project repository or website would be nice on `Web-Token`

Resources/doc/10-web-token.rst
```diff
@@ -0,0 +1,18 @@
...
+    This feature is only available woth PHP 8.1+ and Symfony 6.1+.
+
+
+#. Dependency Installation:
+
+#. Keys and Keysets Creation:
+
+#. BundleConfiguration:
+
+#. Key Rotation:
```
chalasr: Help welcome to fill the docs :)

Command/MigrateConfigCommand.php
```diff
@@ -0,0 +1,230 @@
...
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\DependencyInjection\Compiler\ValidateEnvPlaceholdersPass;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Extension\ConfigurationExtensionInterface;
+use Symfony\Component\DependencyInjection\Extension\ExtensionInterface;
+use Symfony\Component\Yaml\Yaml;
+
+/**
+ * @author Nicolas Cabot <n.cabot@lexik.fr>
```
chalasr: This looks wrong :)

composer.json
```diff
@@ -84,11 +84,11 @@
...
-            "vendor/bin/simple-phpunit",
-            "ENCODER=lcobucci vendor/bin/simple-phpunit",
-            "ENCODER=lcobucci ALGORITHM=HS256 vendor/bin/simple-phpunit",
-            "ENCODER=user_id_claim vendor/bin/simple-phpunit",
-            "PROVIDER=lexik_jwt vendor/bin/simple-phpunit"
+            "vendor/bin/simple-phpunit --exclude-group web-token",
+            "ENCODER=lcobucci vendor/bin/simple-phpunit --exclude-group web-token",
+            "ENCODER=lcobucci ALGORITHM=HS256 vendor/bin/simple-phpunit --exclude-group web-token",
+            "ENCODER=user_id_claim vendor/bin/simple-phpunit --exclude-group web-token",
+            "PROVIDER=lexik_jwt vendor/bin/simple-phpunit --exclude-group web-token"
```
chalasr, line 90: Ideally this composer script should tests all integrations, but I see that you added a dedicated workflow for this one. 
What is the motivation? If it's only to not require(-dev) the necessary packages then no worry, we can and should require(-dev) them. But maybe I'm missing something else?

Tests/Functional/app/AppKernel.php
```diff
@@ -41,12 +41,17 @@ public function __construct($environment, $debug, $testCase = null)
...
     {
-        return [
+        $bundles = [
             new FrameworkBundle(),
             new SecurityBundle(),
             new LexikJWTAuthenticationBundle(),
             new Bundle(),
         ];
+        if (class_exists(JoseFrameworkBundle::class)) {
+            $bundles[] = new JoseFrameworkBundle();
```
chalasr, line 53: If it's always installed, we could remove the class existence check

Command/MigrateConfigCommand.php
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    public function isEnabled(): bool
+    {
+        $requirements = [
+            JoseFrameworkBundle::class => 'web-token/jwt-bundle',
+            JWKFactory::class => 'web-token/jwt-key-mgmt',
+            ClaimCheckerManager::class => 'web-token/jwt-checker',
+            JWEBuilder::class => 'web-token/jwt-encryption',
+            JWELoader::class => 'web-token/jwt-signature',
```
chalasr: probably not the right class as JWELoader is from `web-token/jwt-encryption`

Command/MigrateConfigCommand.php
```diff
@@ -0,0 +1,230 @@
...
+        ];
+
+        foreach (array_keys($requirements) as $requirement) {
+            if (!class_exists($requirement)) {
+                return false;
+            }
+        }
+
+        return true;
```
chalasr: I suggest to rather throw a meaningful exception in `execute()` telling the developer to install the required packages if they are not available, so they can catch the issue when running this command (as they will probably discover the integration in the docs). Btw I'm pretty sure this command should never end up being registered in production, right? Can we remove its service definition in non debug mode maybe?

Events.php
```diff
@@ -81,4 +81,12 @@ final class Events
      * @Event("Lexik\Bundle\JWTAuthenticationBundle\Event\JWTExpiredEvent")
      */
     public const JWT_EXPIRED = 'lexik_jwt_authentication.on_jwt_expired';
+
+    /**
+     * Dispatched before the JWE is computed.
+     * This event allow the JWE header parameters to be changed.
```
chalasr, line 87: A small note telling that this event is only dispatched when using the web-token integration would be nice.

DependencyInjection/LexikJWTAuthenticationExtension.php
```diff
@@ -186,4 +189,37 @@ private static function createTokenExtractors(ContainerBuilder $container, array
...
+            if ($config['access_token_verification']['encryption']['enabled'] === true) {
+                $container->setParameter('lexik_jwt_authentication.web_token.loader.continue_on_decryption_failure', $config['access_token_verification']['encryption']['continue_on_decryption_failure']);
+                $container->setParameter('lexik_jwt_authentication.web_token.loader.jwe_header_checkers', $config['access_token_verification']['encryption']['header_checkers']);
+                $container->setParameter('lexik_jwt_authentication.web_token.loader.allowed_key_encryption_algorithms', $config['access_token_verification']['encryption']['allowed_key_encryption_algorithms']);
+                $container->setParameter('lexik_jwt_authentication.web_token.loader.allowed_content_encryption_algorithms', $config['access_token_verification']['encryption']['allowed_content_encryption_algorithms']);
+                $container->setParameter('lexik_jwt_authentication.web_token.loader.encryption_keyset', $config['access_token_verification']['encryption']['keyset']);
```
chalasr: We should avoid creating parameters and call  `replaceArgument()` on the relevant definitions instead to avoid cluttering the cached container with user-configured values.

Services/WebToken/AccessTokenLoader.php
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * @return array<string, mixed>
+     */
+    private function loadJWS(string $token): array
+    {
+        $payload = null;
+        $data = null;
+        $signature = null;
+        try {
+            $jws = $this->jwsLoader->loadAndVerifyWithKeySet($token, $this->signatureKeyset, $signature);
```
chalasr, line 88: could we find a way to take the configured `clock_skew` into account for time-based claims? If not, we should mention that limitation in the docs as for the token split feature.

chalasr (commented):
First round of comments and questions :)

Services/WebToken/AccessTokenLoader.php
Spomky: OK. This will be done

Services/WebToken/AccessTokenLoader.php
Spomky: Sure. No pb

Tests/Functional/app/AppKernel.php
Spomky: I did that because the current CI includes PHP 7.1, which is not supported by Web-Token (unless out-of-life versions are used). 
This also explains why I excluded the tests in the composer scripts.

Services/WebToken/AccessTokenLoader.php
Spomky: Done

Services/WebToken/AccessTokenLoader.php
Spomky: Done

Resources/doc/index.rst
Spomky: A link is now present

Resources/doc/10-web-token.rst
Spomky: I will write a few words about this feature. I'm just waiting for last minute changes before finishing this part.

Command/MigrateConfigCommand.php
Spomky: Fixed

Command/MigrateConfigCommand.php
Spomky: Reasons given in https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1041#discussion_r958838652

Events.php
Spomky: Note added

DependencyInjection/LexikJWTAuthenticationExtension.php
Spomky: Make sense. I replaced all parameters in favor of `replaceArgument()` calls

DependencyInjection/Configuration.php
```diff
@@ -115,7 +115,120 @@ public function getConfigTreeBuilder(): TreeBuilder
...
-            ->end();
+                ->arrayNode('access_token_issuance')
+                    ->fixXmlConfig('access_token_issuance')
+                    ->canBeEnabled()
+                    ->children()
+                        ->arrayNode('signature')
+                            ->fixXmlConfig('signature')
+                            ->addDefaultsIfNotSet()
+                            ->children()
+                                ->scalarNode('signature_algorithm')
```
chalasr: Can we rename to `algorithm`? `signature_algorithm` feels slightly redundant given it is under the `signature` node. If that works for you, please do the same for all similar nodes

Encoder/WebTokenEncoder.php
```diff
@@ -0,0 +1,65 @@
...
+ * Json Web Token encoder/decoder based on the web-token framework.
+ *
+ * @author Robin Chalas <robin.chalas@gmail.com>
+ */
+class WebTokenEncoder implements HeaderAwareJWTEncoderInterface
+{
+    /**
+     * @var AccessTokenBuilder|null
+     */
+    protected $accessTokenBuilder;
```
chalasr:
```suggestion
    private $accessTokenBuilder;
```

Encoder/WebTokenEncoder.php
```diff
@@ -0,0 +1,65 @@
...
+
+use Lexik\Bundle\JWTAuthenticationBundle\Exception\JWTDecodeFailureException;
+use Lexik\Bundle\JWTAuthenticationBundle\Exception\JWTEncodeFailureException;
+use Lexik\Bundle\JWTAuthenticationBundle\Services\WebToken\AccessTokenBuilder;
+use Lexik\Bundle\JWTAuthenticationBundle\Services\WebToken\AccessTokenLoader;
+
+/**
+ * Json Web Token encoder/decoder based on the web-token framework.
+ *
+ * @author Robin Chalas <robin.chalas@gmail.com>
```
chalasr: not true :)

Encoder/WebTokenEncoder.php
```diff
@@ -0,0 +1,65 @@
...
+        $this->accessTokenLoader = $accessTokenLoader;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function encode(array $payload, array $header = [])
+    {
+        if (!$this->accessTokenBuilder) {
+            throw new \RuntimeException('Not implemented');
```
chalasr: Since this is the result of a developer error that requires changing the code, I suggest to change into to a `\LogicException` with a message that hints about what is misconfigured

Encoder/WebTokenEncoder.php
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function decode($token)
+    {
+        if (!$this->accessTokenLoader) {
+            throw new JWTDecodeFailureException(JWTDecodeFailureException::UNVERIFIED_TOKEN, 'Unable to load the token.');
```
chalasr: This should probably be consistent with `encode()`'s behavior above:  \LogicException with good indication in the message

Encoder/WebTokenEncoder.php
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        try {
+            return $this->accessTokenLoader->load($token);
+        } catch (JWTDecodeFailureException $e) {
+            throw $e;
+        } catch (\Throwable $e) {
```
chalasr: Can we avoid any catch-all and only catch what's expected to be thrown? If something else is thrown, it means the code needs to be fixed thus a 500 response code is legit.

DependencyInjection/LexikJWTAuthenticationExtension.php
```diff
@@ -141,7 +141,7 @@ public function load(array $configs, ContainerBuilder $container)
                 ->replaceArgument(2, $config['public_key'])
                 ->replaceArgument(3, $config['pass_phrase'])
                 ->replaceArgument(4, $encoderConfig['signature_algorithm']);
-            if ($container->getParameter('kernel.debug') !== true) {
+            if (!$container->hasParameter('kernel.debug') || $container->getParameter('kernel.debug') === false) {
```
chalasr:
```suggestion
            if (!$container->hasParameter('kernel.debug') || !$container->getParameter('kernel.debug')) {
```

chalasr (commented):
Some more comments :) I'm really looking forward to merge this. 
Also FYI version the 3.x branch of this bundle is going to require PHP 8.1 minimum, so we will be able to remove the specificities there.

Encoder/WebTokenEncoder.php
Spomky: OK done<

Encoder/WebTokenEncoder.php
Spomky: Both methods throw a LogicException in case of disabled feature. Also, the message was updated

Encoder/WebTokenEncoder.php
Spomky: Coorect. Done

Encoder/WebTokenEncoder.php
Spomky: Fixed

Encoder/WebTokenEncoder.php
Spomky: Done

DependencyInjection/Configuration.php
Spomky: Yes it is redundant. I removed the prefix `signature_` here and also for `allowed_signature_algorithms`

Command/MigrateConfigCommand.php
Spomky: The service definition is now removed unless the debug mode in enabled

Services/WebToken/AccessTokenLoader.php
Spomky: I think this is possible. I need more time for covering this feature

chalasr: It's ok if it's for another PR, people can keep using the lcobucci integration if they have a strong need for it until we implement that feature.

Spomky: Actually it is not difficult at all.
It uses the concept of "claim/header checkers" (validators would be better). By default, it uses the following list: `['exp', 'iat', 'nbf']`.
I just need to tweak thoses aliases to make them `clock_skew` compatible.

Spomky: Addressed through https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1041/commits/c9a33d79802a8dc62cfbe8efde433e1a63f4f3a4

Spomky:
Hi,

I rebased the and finished the last things including the documentation.
If anyone passing here can test it, I'll be happy to receive comments and feedback on it.

Many thanks.

chalasr:
It took time, but here we go, this is in now. Thank you very much @Spomky.

Spomky:
Excellent! TBH I missed this PR and forget to have it in line with the branch.
Let me know if there is any trouble with this feature and I will take care of it.

PR #1043, opened 2022-06-29
Fix deprecation message

mbabker:
Right now, the message this is generating says `Since lexik/jwt-authentication-bundle 2.5: [snip] since version 2.15`, clearly showing two different versions.  The deprecation actually landed in 2.16, so this updates the message to use a single (and consistent) version number.


chalasr:
Oops, bad c/p. Good catch!

chalasr:
Thank you Michael.

PR #1060, opened 2022-08-24
Fix CI

chalasr:


misantron:
`lcobucci/jwt` 4.2.0 contains BC, for example: https://github.com/lcobucci/jwt/pull/835/files. Now all attempts to generate JWT with symmetric algorithms and small length keys will fail.

Issue #1066, opened 2022-09-06
TypeError in JWTManager


magikid:
# What happened

The token extractor failed to extract the token from the request and threw a TypeError on the `JWTManager::parse(string $token)`.

# What I expected

I expected an exception to be thrown about a malformatted token.

# Other notes

It looks when extracting the token in [JWTAuthenticator](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/d7cbef600712eb57af7cafd28130985553b50b37/Security/Authenticator/JWTAuthenticator.php) the `extract(Request $request)` function returns `string|false`.  When that function returns `false`, it is stored in the `$token` variable and passed to the `JWTManager::parse(string $token)` function which can't accept a boolean.

magikid:
This has been merged already. Closing issue.

PR #1067, opened 2022-09-06
Fixes TypeError in JWTManager

magikid:
When the TokenExtractor extracts a token, it will return
`string|false`.  The JWTAuthenticator then passes this value to
`JWTManager::doAuthenticate` which can only accept a string.  If a false
is returned by the TokenExtractor, PHP throws a type error.  This commit
checks the return value and throws an exception if it returns false.

Issue: #1066

### DIFF
--- a/Security/Authenticator/JWTAuthenticator.php
+++ b/Security/Authenticator/JWTAuthenticator.php
@@ -105,7 +105,7 @@ class JWTAuthenticator extends AbstractAuthenticator implements AuthenticationEn
     {
         $token = $this->getTokenExtractor()->extract($request);
         if ($token === false) {
-            throw new InvalidTokenException('Unable to extract JWT token');
+            throw new \LogicException('Unable to extract a JWT token from the request. Also, make sure to call `supports()` before `authenticate()` to get a proper client error.');
         }
 
         try {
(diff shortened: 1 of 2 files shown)


Security/Authenticator/JWTAuthenticator.php
```diff
@@ -104,6 +104,9 @@ public function supports(Request $request): ?bool
     public function doAuthenticate(Request $request) /*: Passport */
     {
         $token = $this->getTokenExtractor()->extract($request);
+        if ($token === false) {
+            throw new InvalidTokenException('Unable to extract JWT token');
```
chalasr: This should be a `\LogicException` to me as it is about a developer error and as such require fixing the calling code. Method `authenticate()` should not be called unless `supports()` returned true, which prevents reaching this case.

chalasr:
```suggestion
            throw new \LogicException('Unable to extract a JWT token from the request. Also, make sure to call `supports()` before `authenticate()` to get a proper client error.');
```

Tests/Security/Authenticator/JWTAuthenticatorTest.php
```diff
@@ -307,6 +307,24 @@ public function testCreateAuthenticatedToken()
...
+            ->willThrowException(new InvalidTokenException('Unable to extract JWT token'));
+
+        $authenticator = new JWTAuthenticator(
+            $jwtManager,
+            $this->getEventDispatcherMock(),
+            $this->getTokenExtractorMock(false),
+            $this->getUserProviderMock()
+        );
+
+        $this->expectException(InvalidTokenException::class);
```
chalasr:
```suggestion
        $this->expectException(\LogicException::class);
```

chalasr:
Thank you @magikid.

PR #1074, opened 2022-10-09
Update link to refresh token bundle

Nek-:
Well, the old link was correctly redirected to the new repository. But let's fix it before the redirection ends :) .


chalasr (approved):

chalasr:
Good catch, thanks Maxime.

PR #1075, opened 2022-10-10
Support an attribute to answer yes when generating keys

### LINKED ISSUE #1073: Support an attribute to answer yes when generating keys
I wish to be able to use `lexik:jwt:generate-keypair` command with `--overwrite` and no promt, just generate it. I haven't found a way so far and a `-y` argument is pretty standard.

vasilvestre:
Fix https://github.com/lexik/LexikJWTAuthenticationBundle/issues/1073

### DIFF
--- a/Command/GenerateKeyPairCommand.php
+++ b/Command/GenerateKeyPairCommand.php
@@ -123,10 +123,12 @@ final class GenerateKeyPairCommand extends Command
                 return 1;
             }
 
-            if (null === $input->getOption('yes') && !$io->confirm('You are about to replace your existing keys. Are you sure you wish to continue?')) {
-                $io->comment('Your action was canceled.');
-
-                return 0;
+            if (!$input->getOption('yes')) {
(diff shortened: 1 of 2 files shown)


chalasr:
@vasilvestre Do you plan to finish this PR at some point? Or are you looking for some help?

vasilvestre:
> @vasilvestre Do you plan to finish this PR at some point? Or are you looking for some help?

I'm not sure the option is useful, passing -n does the same AFAIK. Should try first..
Wdyt ?

chalasr:
If `--overwrite` already does the trick as it seems to do, I don't think we need a new option. At best a shortcut for this option (I thought it was the purpose of the linked issue)

vasilvestre:
> If `--overwrite` already does the trick as it seems to do, I don't think we need a new option. At best a shortcut for this option (I thought it was the purpose of the linked issue)

I couldn't find the -n option at this moment. I'm closing it so

PR #940, opened 2021-10-31
Add `remove_token_from_body_when_cookies_used` config option

TjorvenB:
Added configuration and documentation for the argument added in Lexik#912. https://github.com/lexik/LexikJWTAuthenticationBundle/pull/912

PR #1092, opened 2022-12-03
allow environment variables for `remove_token_from_body_when_cookies_used`

usu:
https://github.com/lexik/LexikJWTAuthenticationBundle/pull/940 introduced a new config option to enable response body when cookies are used:

```yaml
remove_token_from_body_when_cookies_used: false
```

However, currently environment variables cannot be used to control this config. An error `Environment variables "bool:JWT_REMOVE_TOKEN_FROM_BODY" are never used` is thrown. After this PR, this option can also be used as following:

```yaml
remove_token_from_body_when_cookies_used: '%env(bool:JWT_REMOVE_TOKEN_FROM_BODY)%'
```

### DIFF
--- a/DependencyInjection/LexikJWTAuthenticationExtension.php
+++ b/DependencyInjection/LexikJWTAuthenticationExtension.php
@@ -102,10 +102,10 @@ class LexikJWTAuthenticationExtension extends Extension
             ->getDefinition('lexik_jwt_authentication.extractor.chain_extractor')
             ->replaceArgument(0, $tokenExtractors);
 
-        if (false === $config['remove_token_from_body_when_cookies_used']) {
+        if (isset($config['remove_token_from_body_when_cookies_used'])) {
             $container
                 ->getDefinition('lexik_jwt_authentication.handler.authentication_success')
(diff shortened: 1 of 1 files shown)


chalasr:
Thanks for the PR. Did you actually test it? I suspect this is not enough to properly support setting it via an env variable:
https://github.com/symfony/symfony/blob/043257f6fc1fceb28bc0c37c59f2bea4d9c8d79d/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L1945

usu:
I tested successfully with my own application. Symfony DI automatically resolves the environment parameters when generating the DI factories. I assume this is the reason why environment variables already work for all the other config options.

Happy to provide a test, but couldn't find a good test to hook into or to copy from. There was no test provided when introducing #940 and couldn't really find a test that already utilizes environment variables for other config options.

usu:
Hi @chalasr, anything you'd want me to add to this PR to make this mergeable?

chalasr (approved):

chalasr:
Thank you @usu.

PR #1093, opened 2022-12-06
Fixed: Allow expiration never (ttl = 0)

Ninos:
Setting `$payload['exp'] = 0;` in JWT_CREATED event, the exp entry is removed correctly from payload, but the validation of token is always returning invalid. If no exp is defined, the exp is unlimited by jwt-standard. This PR should fix it.


Ninos:
Nvm, I haven't seen the setting `allow_no_expiration: true`. Closed.

PR #1101, opened 2022-12-20
Support Unsafe signer algorithms

dmitryuk:
Since https://github.com/lcobucci/jwt/releases/tag/4.2.0 some signer algorithms requires specific key length, for example, Sha256 requires > 32 symbols. 
But my company has shorter keys, which we can't change because we have dozens of services. 
The PR adds possibility to use deprecated signers


maxhelias:
I am not in favor of adding this as they are currently deprecated and are removed in version 5

chalasr:
Closing for the reason given by @maxhelias. PR welcome for supporting the v5 btw :)

PR #1110, opened 2023-02-02
Use the Security domain for translated messages

jderusse:
By default, the translation keys related to security are stored in the `security` domain.

The AuthenticationFailureHandler use this domain:

https://github.com/lexik/LexikJWTAuthenticationBundle/blob/2.x/Security/Http/Authentication/AuthenticationFailureHandler.php#L46

But these 2 authenticators don't use the right domain.


chalasr (approved):

maxhelias (approved):

chalasr:
Thank you Jérémy.

PR #1112, opened 2023-02-07
fix: better API Platform and json_login compatibility

alanpoulain:
After https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1098.

This PR adds two new arguments for the API Platform compatibility: `username_path` and `password_path`.

Moreover, it is not necessary anymore to add `check_path` to enable the decoration of the OpenAPI factory: if API Platform is detected and if a `json_login` firewall is available, it is automatically enabled!

For complex configuration, the user can take over the configuration.


chalasr (approved):
Awesome!

vincentchalamon (approved):

chalasr:
Thank you @alanpoulain.

PR #1108, opened 2023-01-27
Clean 3.x

maxhelias:
- [x] Remove namshi/jose
- [x] Remove support < PHP8.1 & < Symfony 5.4
- [x] Support of PHPUnit 9.5
- [x] Update documentation
- [x] Create an upgrade file

PR #1115, opened 2023-02-10
Fix ci 2.x

### LINKED ISSUE #1114: v2.18.0 Incompatible with lcobucci/jwt ^3.4
#1109 introduced `LooseValidAt` constraint from [lcobucci/jwt](https://github.com/lcobucci/jwt), but the class only exists on `4.0.x` and above, but the composer.json states that the bundle is compatible with: `lcobucci/jwt: "^3.4|^4.0"`.

Installing `lcobucci/jwt:^3.4` with `lexik/jwt-authentication-bundle:v2.18.0` results in error:
`Error: Class "Lcobucci\JWT\Validation\Constraint\LooseValidAt" not found`

Possible solutions:
- Check if `LooseValidAt` class is available, if not use  ` ValidAt` class.
- Make this bundle only compatible with `lcobucci/jwt:^4.0`

maxhelias:
Fix https://github.com/lexik/LexikJWTAuthenticationBundle/issues/1114 and all CI.

PHP 7.1 and symfony 4.4 cannot be tested in the CI but are still supported. Mainly because of the tests that need the browser-kit 5.4 at least and interface compatibility (without/with typing)

@chalasr after merge this, I would like a rebase on 3.x so that I can rework my other PR : https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1108

### DIFF
--- a/Services/JWSProvider/LcobucciJWSProvider.php
+++ b/Services/JWSProvider/LcobucciJWSProvider.php
@@ -23,6 +23,7 @@ use Lcobucci\JWT\Token\Plain;
 use Lcobucci\JWT\Token\RegisteredClaims;
 use Lcobucci\JWT\Validation\Constraint\LooseValidAt;
 use Lcobucci\JWT\Validation\Constraint\SignedWith;
+use Lcobucci\JWT\Validation\Constraint\ValidAt;
 use Lcobucci\JWT\Validation\Validator;
 use Lcobucci\JWT\ValidationData;
 use Lexik\Bundle\JWTAuthenticationBundle\Services\KeyLoader\KeyLoaderInterface;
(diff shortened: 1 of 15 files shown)


Services/JWSProvider/LcobucciJWSProvider.php
```diff
@@ -259,10 +264,11 @@ private function verify(Token $jwt)
         }
 
         $validator = new Validator();
+        $classValidator = class_exists(LooseValidAt::class) ? LooseValidAt::class : ValidAt::class;
```
maxhelias, line 263: Fix compatibility with lcobucci/jwt 3.4 and 4.x

Command/CheckConfigCommand.php
```diff
@@ -3,6 +3,7 @@
 namespace Lexik\Bundle\JWTAuthenticationBundle\Command;
 
 use Lexik\Bundle\JWTAuthenticationBundle\Services\KeyLoader\KeyLoaderInterface;
+use RuntimeException;
```
chalasr: Sorry to ask but can you please revert this change? I prefer the current approach as it makes it easy to distinguish built-in exceptions from domain ones

maxhelias: Yes, I also need to find the rules rector to disable this change

maxhelias: Done 😉

chalasr:
Thanks for fixing this bug @maxhelias.

webhdx:
Hey @chalasr. When do you plan to make 2.18.1 with this fix?

chalasr:
@webhdx Here you go https://github.com/lexik/LexikJWTAuthenticationBundle/releases/tag/v2.18.1

PR #1119, opened 2023-02-20
Fix API Platform integration

### LINKED ISSUE #1118: Deactivate API Platform automatic integration
Hello,

Since the version [2.18.0](https://github.com/lexik/LexikJWTAuthenticationBundle/releases/tag/v2.18.0) the API Platform integration is automatically enabled if the bundle detects that API Platform is installed.

This is a great addition and should work for most projects, but can also cause some issue is some specific case. For example I'm using split cookies and have handcrafted my own definition for the API login operation. So after the update, I now have in the documentation page two definitions for the login route: mine and the one autogenerated (with a wrong definition).

Is there a way to deactive the automatic integration (I didn't find any) or would you consider a PR to allow that? If so, what whould be the best approach? Adding a new parameter under `lexik_jwt_authentication.api_platform` or something else?

Thank you!

maxhelias:
Fix https://github.com/lexik/LexikJWTAuthenticationBundle/issues/1118

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -2,6 +2,7 @@
 
 namespace Lexik\Bundle\JWTAuthenticationBundle\DependencyInjection;
 
+use ApiPlatform\Symfony\Bundle\ApiPlatformBundle;
 use Symfony\Component\Config\Definition\BaseNode;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
@@ -116,6 +117,7 @@ class Configuration implements ConfigurationInterface
                     ->end()
                 ->end()
                 ->arrayNode('api_platform')
+                    ->canBeEnabled()
                     ->info('API Platform compatibility: add check_path in OpenAPI documentation.')
                     ->children()
                         ->scalarNode('check_path')
--- a/DependencyInjection/LexikJWTAuthenticationExtension.php
+++ b/DependencyInjection/LexikJWTAuthenticationExtension.php
@@ -99,7 +99,7 @@ class LexikJWTAuthenticationExtension extends Extension
         $container->setParameter('lexik_jwt_authentication.encoder.signature_algorithm', $encoderConfig['signature_algorithm']);
         $container->setParameter('lexik_jwt_authentication.encoder.crypto_engine', $encoderConfig['crypto_engine']);
 
-        $tokenExtractors = self::createTokenExtractors($container, $config['token_extractors']);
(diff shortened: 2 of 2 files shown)


alanpoulain:
Couldn't we add an enabled parameter (false by default, but true in the recipe)? It's too bad to give up this behavior.

maxhelias:
@alanpoulain I can modify to add `canEnabled()` in the configuration but having it true in the recipe without APIP installed may break the bundle out of the box

maxhelias:
But it could be set to true with a prepend in the APIP extension, no ?

alanpoulain:
Having it to true should not prevent the detection of API Platform. This parameter is only to disable the integration if needed and prevent the BC break.
Doing it with a prepend in API Platform would be a BC break too.

maxhelias:
@alanpoulain should be good now
@chalasr the CI "Tests / PHP 8.2 - Sf 6.3" should be fix in https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1120

DependencyInjection/Configuration.php
```diff
@@ -116,6 +117,7 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->end()
                 ->end()
                 ->arrayNode('api_platform')
+                    ->{class_exists(ApiPlatformBundle::class) ? 'canBeDisabled' : 'canBeEnabled'}()
```
alanpoulain: Isn't the BC break still here? IMO it should be:
```suggestion
                    ->canBeEnabled()
```

maxhelias: It shouldn't, in this way we don't need to put it to true in the recipe and won't throw the exception if APIP is not installed (except if defined by the user in the recipe)

alanpoulain: But it doesn't solve https://github.com/lexik/LexikJWTAuthenticationBundle/issues/1118: this person is using API Platform and has upgraded to LexikJWT 2.18. Their BC break is that an OpenAPI route has been added just by upgrading.

chalasr: Agree, defaulting to `false` makes sense to me. A new version of the recipe enabling it feels less aggressive as it will only be installed for new projects and recipe upgrades.

chalasr (approved):

jonag:
Thank you for your PR, it should fix the issue described in #1118. Personally I'm fine with both options (enabled ou disabled by default) since (in my case at least) it doesn't "break" my application but just add a new (non-working and wrongly documented) operation. I don't mind having to disable it manually.

chalasr:
Thank you @maxhelias.

PR #1120, opened 2023-02-20
Remove deprecation symfony 6.3

maxhelias:
This is to remove deprecation notices on the Symfony 6.3 CI and adding this type should not break the bundle because all of them are on internal elements

### DIFF
--- a/Command/CheckConfigCommand.php
+++ b/Command/CheckConfigCommand.php
@@ -36,7 +36,7 @@ class CheckConfigCommand extends Command
     /**
      * {@inheritdoc}
      */
-    protected function configure()
+    protected function configure(): void
     {
         $this
             ->setName(static::$defaultName)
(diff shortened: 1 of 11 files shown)


chalasr (approved):
Thanks!

LaYoance:
Thanks ! 

Inside the Lexik\Bundle\JWTAuthenticationBundle\Security\Authenticator

Symfony\Component\Security\Http\Authenticator\Passport\PassportInterface does not exists anymore since SF 6.0
You should use directly Symfony\Component\Security\Http\Authenticator\Passport\Passport instead

maxhelias:
@LaYoance Yes but it's only used for the compatibility of `createAuthenticatedToken` method. This is not a problem in this configuration, but if we change it, it will break the compatibility with the interface.

And this method will be removed in the near future because it is depreciated.

LaYoance:
Well, I understand that.

An old class in our project was using it. I've updated it on my side to match the new one.

Thanks

chalasr:
Good catch, thanks Maxime.

mickverm:
Symfony 6.3.0-BETA1 has been released, any chance this change could also make it to a release?

PR #1125, opened 2023-03-01
Allow lcobucci/jwt v5

maxhelias:


maxhelias:
Humm lcobucci/jwt need php:^8; Maybe we can do this for version 3?

chalasr:
How hard is it to support both v3 and v5? Ideally we'd let composer do its resolution job

maxhelias:
@chalasr Actually, no, it's okay 😅

chalasr:
Awesome!

chalasr:
Thank you @maxhelias.

Issue #996, opened 2022-03-17
Fatal error Cannot declare trait ForwardCompatAuthenticatorTrait


rpilote:
I got and error after install JWT package 

<br />
<b>Fatal error</b>: Cannot declare trait
Lexik\Bundle\JWTAuthenticationBundle\Security\Authenticator\ForwardCompatAuthenticatorTrait, because the name is already
in use in
<b>/srv/api/vendor/lexik/jwt-authentication-bundle/Security/Authenticator/ForwardCompatAuthenticatorTrait.php(13) : eval()'d code</b>
on line <b>10</b><br />

Is some one can't help me I can't found any clue why i gote this error 

i try with php 8.0.2 and php 8.1 symfony 6.0  and both does work

<img width="1068" alt="image" src="https://user-images.githubusercontent.com/94406729/158910875-c35337ce-f9cd-4d77-bfc6-9ef703bc0151.png">

yivi:
This seems related to preloading. I started getting this recently after some upgrades, and I disabled preloading temporarily until I figured out what's going on.

The issue seems to happen with preload activated and Symfony 6+.

On Symfony 5.4, all good.

rpilote:
I start a new projet and i still got the error. 

this is my security.yaml :
# api/config/packages/security.yaml
security:
    # https://symfony.com/doc/current/security.html#c-hashing-passwords
    password_hashers:
        App\Entity\UserAccount: 'auto'

    # https://symfony.com/doc/current/security/authenticator_manager.html
    enable_authenticator_manager: true
    # https://symfony.com/doc/current/security.html#where-do-users-come-from-user-providers
    providers:
        # used to reload user from session & other features (e.g. switch_user)
        app_user_provider:
            entity:
                class: App\Entity\UserAccount
                property: username 

    firewalls:
        dev:
            pattern: ^/_(profiler|wdt)
            security: false
            
        api:
            pattern: ^/api/
            stateless: true
            provider: app_user_provider
            jwt: ~
            
        main:
            stateless: true
            provider: app_user_provider
            json_login:
                check_path: /authentication_token
                username_path: username
                password_path: password
                success_handler: lexik_jwt_authentication.handler.authentication_success
                failure_handler: lexik_jwt_authentication.handler.authentication_failure

    access_control:
        - { path: ^/docs, roles: PUBLIC_ACCESS } # Allows accessing the Swagger UI
[...]

PR #1134, opened 2023-03-30
Fix ForwardCompatAuthenticatorTrait with OPCache preload

elavrom:
Hi there,

Since the `eval()` in the `ForwardCompatAuthenticatorTrait` is creating problems when OPCache preload is enabled, here is a quick fix to avoid eval()'ing multiple times.

(Issue ref : #996)

P.S. : First time making a PR, feel free to tell me if I did something wrong

### DIFF
--- a/Security/Authenticator/ForwardCompatAuthenticatorTrait.php
+++ b/Security/Authenticator/ForwardCompatAuthenticatorTrait.php
@@ -9,8 +9,9 @@ use Symfony\Component\Security\Http\Authenticator\Passport\PassportInterface;
 
 $r = new \ReflectionMethod(AuthenticatorInterface::class, 'authenticate');
 
-if ($r->hasReturnType() && Passport::class === $r->getReturnType()->getName()) {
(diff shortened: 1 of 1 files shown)


Amegatron:
I still wonder, why there is need for `eval`. It seems to be OK to have just normal Trait declarations. Instead, eval indeed may cause problems with different kinds of static analysis and caching.

chalasr:
The reason is that the v2 of this package supports a wide range of PHP versions including legacy ones that do not support some type declarations. Using `eval` is the only way to make php skip interpreting it and avoid ending up with a fatal error.

chalasr:
Thank you @elavrom.

Amegatron:
@chalasr Thanks for clarification. I believe this reason now especially highlights what I described in the mentioned issue: https://github.com/lexik/LexikJWTAuthenticationBundle/issues/996#issuecomment-1501530540

In brief, my belief is that another compatibility approach should be chosen.

PR #1144, opened 2023-06-05
Fix missing array claims BC break in 2.9.0

### LINKED ISSUE #950: Why is SelfValidatingPassport imposed in JWTAuthenticator?
Hello,

Still trying to extend the new `JWTAuthenticator` in my system.

In my JWT, I have additional data referencing a `Session` object we have and I need to check whether this session is still active when trying to login (this allows forced logout in case of security breaches). 

I migrated this logic into the new Authenticator system by providing a `CustomCredentials` to a simple `Passport` in my `::authenticate()` method.

But I'm then faced with the fact that `JWTAuthenticator::createAuthenticatedToken` checks that the passport receives is a `SelfValidatingPassport` and not simply a `Passport` or `PassportInterface`.

This forces me to override the method, and add the `EventDispatcher` as a property of my authenticator to be able to dispatch manually the `JWTAuthenticatedEvent`.

This seems really restrictive, as it prevents us from using custom credentials with the default `JWTAuthenticator::createAuthenticatedToken` method.

Is there a specific reason to be so restrictive that I'm not aware of?

`AbstractAuthenticator::createAuthenticatedToken` already checks that the `$passport` is a `UserPassportInterface`.

### LINKED ISSUE #1139: aud property is missing in 2.19
We add `aud` to payload via `lexik_jwt_authentication.on_jwt_created` event. After upgrade to 2.19, this property is missing in encoded token

Our listener
```php
<?php

use Lexik\Bundle\JWTAuthenticationBundle\Event\JWTCreatedEvent;
use Lexik\Bundle\JWTAuthenticationBundle\Events;
use Symfony\Component\EventDispatcher\EventSubscriberInterface;

final class AttachPayloadToJwtSubscriber implements EventSubscriberInterface
{
    public function __construct(private PlatformDetector $platformDetector)
    {
    }

    public function onJwtCreated(JWTCreatedEvent $event): void
    {
        /** @var User $user */
        $user = $event->getUser();
        $payload = ['aud' => [],];

        foreach ($user->getLicense()->getApplications() as $licenseApplication) {
            $payload['aud'][] = $licenseApplication->getApplication()->getId();
        }

        $event->setData(array_merge($event->getData(), $payload));
    }

    public static function getSubscribedEvents(): array
    {
        return [Events::JWT_CREATED => 'onJwtCreated'];
    }
}
```

ostrolucky:
Regression from 02a7bc03cf545695c086f300d58e5610d5888871 
Fixes #1139

### DIFF
--- a/Services/JWSProvider/LcobucciJWSProvider.php
+++ b/Services/JWSProvider/LcobucciJWSProvider.php
@@ -271,7 +271,7 @@ class LcobucciJWSProvider implements JWSProviderInterface
             unset($payload[$claim]);
 
             if (\is_array($value)) {
-                \call_user_func_array([$builder, $mutator], $value);
+                $builder = \call_user_func_array([$builder, $mutator], $value);
                 continue;
             }
 
--- a/Tests/Functional/GetTokenTest.php
+++ b/Tests/Functional/GetTokenTest.php
@@ -58,7 +58,7 @@ class GetTokenTest extends TestCase
 
         $subscriber = static::$kernel->getContainer()->get('lexik_jwt_authentication.test.jwt_event_subscriber');
         $subscriber->setListener(Events::JWT_CREATED, function (JWTCreatedEvent $e) {
-            $e->setData($e->getData() + ['custom' => 'dummy']);
+            $e->setData($e->getData() + ['custom' => 'dummy', 'aud' => ['foo', 'bar']]);
             $e->setHeader($e->getHeader() + ['foo' => 'bar']);
         });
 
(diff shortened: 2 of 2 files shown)


chalasr (approved):

chalasr:
Thanks a lot @ostrolucky, that's much appreciated.

ostrolucky:
Hey not sure when do you plan to do a next release?

chalasr:
@ostrolucky done https://github.com/lexik/LexikJWTAuthenticationBundle/releases/tag/v2.19.1. Sorry for the delay

PR #1156, opened 2023-09-04
feat: drop explizit support of symfony ^5.3. Updated to LTS ^5.4

Chris53897:
I suggest to bump the minimum required version from Symfony 5.3 to 5.4 LTS.
https://github.com/lexik/LexikJWTAuthenticationBundle/blob/2.x/.github/workflows/ci.yml#L17

As there are no Tests for 4.4.
Can we add testruns for 4.4 or better drop support for 4.4 (as it is no longer LTS) and require >=7.2.5 https://github.com/symfony/symfony/blob/5.4/composer.json#L35 ?


chalasr:
I prefer to do nothing until support actually appears to be broken. Thanks for proposing @Chris53897

Issue #1152, opened 2023-08-10
Unrecognized option "enabled" under "lexik_jwt_authentication.api_platform"


thibaut22200:
I have this error message when cache clear in symfony app (version 6.2) :
Executing script cache:clear [KO]
 [KO]
Script cache:clear returned with error code 1
!!
!!  In ArrayNode.php line 304:
!!
!!    Unrecognized option "enabled" under "lexik_jwt_authentication.api_platform"
!!    . Available options are "check_path", "password_path", "username_path".
!!
!!
!!
Script @auto-scripts was called via post-install-cmd

Here is my lexik configuration file:
lexik_jwt_authentication:
    secret_key: '%env(resolve:JWT_SECRET_KEY)%'
    public_key: '%env(resolve:JWT_PUBLIC_KEY)%'
    pass_phrase: '%env(JWT_PASSPHRASE)%'
    token_ttl: 86400

HuyTaing:
Hello there,

I have the same error with a new symfony 6.3 projet.

```
[KO]
Script cache:clear returned with error code 1
!!
!!  In ArrayNode.php line 327:
!!
!!    Unrecognized option "enabled" under "lexik_jwt_authentication.api_platform". Available options are "check_path", "password_path", "username_path".
!!
!!
!!  2023-08-11T09:29:21+00:00 [info] User Deprecated: Method "Symfony\Component\HttpKernel\Bundle\Bundle::build()" might add "void" as a native return type declaration in the future. Do the same in child class "Lexik\Bundle\JWTAuthenticationBundle\LexikJWTAuthenticationBundle" now to avoid errors or add an explicit @return annotation to suppress this message.
!!  2023-08-11T09:29:21+00:00 [info] User Deprecated: Method "Symfony\Component\HttpKernel\Bundle\Bundle::registerCommands()" might add "void" as a native return type declaration in the future. Do the same in child class "Lexik\Bundle\JWTAuthenticationBundle\LexikJWTAuthenticationBundle" now to avoid errors or add an explicit @return annotation to suppress this message.
!!
Script @auto-scripts was called via post-update-cmd
```

there is a problem with lexik/jwt version.
i use composer require lexik/jwt-authentication-bundle by default.

composer.json
```
{
    "type": "project",
    "license": "proprietary",
    "minimum-stability": "stable",
    "prefer-stable": true,
    "require": {
        "php": ">=8.1",
        "ext-ctype": "*",
[...]

PR #1157, opened 2023-09-04
feat: 3.x migrate to "lcobucci/jwt" v5

Chris53897:
feat: add support for psr/clock

To be honest i do not have much knowledge about this topic.

I just followed this mirgration guide https://lcobucci-jwt.readthedocs.io/en/latest/upgrading/
V5 does use [psr/clock](https://packagist.org/packages/psr/clock), instead of v4 with there own clock implementation.

As this is a new major version, i see no need to support versions <5

I added this PR for 3.x Branch as well. But could be a seperate PR if you want to.
https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1155

### DIFF
--- a/Services/JWSProvider/LcobucciJWSProvider.php
+++ b/Services/JWSProvider/LcobucciJWSProvider.php
@@ -2,14 +2,10 @@
 
 namespace Lexik\Bundle\JWTAuthenticationBundle\Services\JWSProvider;
 
-use Lcobucci\Clock\Clock;
-use Lcobucci\Clock\SystemClock;
 use Lcobucci\JWT\Builder;
 use Lcobucci\JWT\Encoding\ChainedFormatter;
 use Lcobucci\JWT\Encoding\JoseEncoder;
-use Lcobucci\JWT\Parser;
 use Lcobucci\JWT\Signer;
-use Lcobucci\JWT\Signer\Ecdsa;
 use Lcobucci\JWT\Signer\Hmac;
 use Lcobucci\JWT\Signer\Hmac\Sha256;
 use Lcobucci\JWT\Signer\Hmac\Sha384;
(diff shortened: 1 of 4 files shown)


Chris53897:
First run failed because of missing `psr/clock` implementation. I just added this to the dev section and let the users decide, which to use. I am not sure for runtime checks if the users to not have an implementation installed.

PHP 8.2 failed because of https://github.com/lexik/LexikJWTAuthenticationBundle/issues/1152
PHP 8.3 fails because [lcobucci/jwt](https://packagist.org/packages/lcobucci/jwt) does not allow PHP 8.3 at the moment.

Chris53897:
Can the fix from https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1119/files or 2.x merged upstream for 3.x?

Chris53897:
I just noticed that @maxhelias used an different approach for 2.x https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1125

maxhelias:
Hi @Chris53897, 2.x should be merge on 3.x to correct compatibility

Chris53897:
After that is done, we could discuss what are the best options.

- Keep support of version 3 and 4?
- Use jwt/clock as hard requirement or let the user decide which implementation of psr/clock the use?

chalasr:
@Chris53897 Thanks for the PR, I'm gonna review and merge soon. I'd plan to finish and release v3.0 somewhere around 2024' first quarter.
Also I'd like to backport this to the 2.x branch with backward compatiblity layers for the older versions of lcobucci/jwt that 2.x supports today. If you feel brave enough to do so, additional PR welcome :)

Chris53897:
Thanks. The chance for additional PRs for the tasks is not so high. My actual main focus is on PRs to support symfony 7 and php 8.3 for repos.

chalasr:
That's fine, thanks for telling me.

composer.json
```diff
@@ -53,6 +53,7 @@
     "require-dev": {
         "phpunit/phpunit": "^9.5",
         "symfony/browser-kit": "^5.4|^6.0",
+        "symfony/clock": "^6.3",
```
stof, line 56: as you instantiate a Symfony Clock when no clock is provided, this is not really an optional dependency.

Services/JWSProvider/LcobucciJWSProvider.php
```diff
@@ -49,10 +47,10 @@ class LcobucciJWSProvider implements JWSProviderInterface
     /**
      * @throws \InvalidArgumentException If the given crypto engine is not supported
      */
-    public function __construct(KeyLoaderInterface $keyLoader, string $signatureAlgorithm, ?int $ttl, ?int $clockSkew, bool $allowNoExpiration = false, Clock $clock = null)
+    public function __construct(KeyLoaderInterface $keyLoader, string $signatureAlgorithm, ?int $ttl, ?int $clockSkew, bool $allowNoExpiration = false, ClockInterface $clock = null)
```
stof, line 50: You should also update the service definition to inject the `clock` service defined by FrameworkBundle when it is available (using `on-invalid="null"` to handle cases where the service is disabled) to use the clock configured in the container.

composer.json
Chris53897: Thanks for the feedback.

I am not sure if this should be tied to "symfony/clock" or if it is possible to just add the requirement that it needs a repo that implements "psr/clock" and let the user decide, which one to chose. require [psr/clock-implementation](https://packagist.org/providers/psr/clock-implementation)?
https://packagist.org/?query=clock

The linked PR from @Dean151 makes the dependency optional.

stof: Well, your implementation is the one making it non-optional.

However, as this package is a Symfony bundle, I would not be too concerned about being bring a dependency on the Symfony component.

PR #1159, opened 2023-09-22
Added: JWT security scheme to openapi

Ninos:
After that it's possible to add token in swagger ui without defining `Bearer` prefix.

PS: Currently there's a bug with `$checkPath`, `$usernamePath` & `$passwordPath`. If you're using route names instead of paths, in swagger ui/docs you get `POST auth` instead of `POST /auth`, so you cannot validate via swagger ui. I haven't found an elegant way to load `GeneratorInterface` service :/

### DIFF
--- a/OpenApi/OpenApiFactory.php
+++ b/OpenApi/OpenApiFactory.php
@@ -44,6 +44,17 @@ class OpenApiFactory implements OpenApiFactoryInterface
         $openApi = ($this->decorated)($context);
 
         $openApi
+            ->getComponents()->getSecuritySchemes()->offsetSet(
+                'JWT',
+                new \ArrayObject([
+                        'type' => 'http',
+                        'scheme' => 'bearer',
+                        'bearerFormat' => 'JWT',
+                    ]
+                )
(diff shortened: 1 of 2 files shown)


Ninos:
Nvm stupid mistake, my firewall-pattern was something like that:
```yaml
security:
    ...
    firewalls:
        dev:
            pattern: ^/(_(profiler|wdt)|css|js|->entityPathName<-)/
```

Ninos:
Sry, fixed functional tests & rebased to latest 2.x branch, hope all's fine now :-)

chalasr (approved):

chalasr:
Thank you @Ninos.

chalasr:
>  Currently there's a bug with $checkPath, $usernamePath & $passwordPath. If you're using route names instead of paths, in swagger ui/docs you get POST auth instead of POST /auth, so you cannot validate via swagger ui. I haven't found an elegant way to load GeneratorInterface service :/

Opening an issue for this with some reproducer would be very nice.

PR #1167, opened 2023-11-13
feat: add support for partitioned cookies

### LINKED ISSUE #1166: Support for partitioned authentication cookies
Hey folks!

With the upcoming [phasing-out of third-party cookies by Chromium](https://developer.chrome.com/docs/privacy-sandbox/third-party-cookie-phase-out/) (and, I expect, other browser vendors in the near future), I think it would be useful to add support for Partitioned authentication cookies in this library.

Some applications embed widgets (or even their entire application!) on other sites in a legitimate way. In these cases, their authentication cookies are usually set with `SameSite=None`, and considered as third-party cookies by browsers. The upcoming phase-out will therefore break any such application... unless they switch to partitioned cookies (see Google's docs on [CHIPS](https://developer.chrome.com/docs/privacy-sandbox/chips/) - _Cookies Having Independent Partitioned State_).

[Symfony has added support for Partitioned Cookies in version 6.4](https://symfony.com/blog/new-in-symfony-6-4-chips-cookies) (which will be released by the end of the month). This means that this library could fairly easily be updated to also allow this option, by simply adding an additional configuration option, which defaults to the current behavior (not partitioned).

Here is what an example configuration for an application in this context could look like, once this is implemented:

```yaml
    set_cookies:
        jwt_hp:
            lifetime: null
            samesite: none
            path: /
            domain: null
            httpOnly: false
            partitioned: true       # <--- new option
            split:
                - header
                - payload

        jwt_s:
            lifetime: 0
            samesite: none
            path: /
            domain: null
            httpOnly: true
            partitioned: true       # <--- new option
            split:
                - signature
```

If the maintainers do not have any objections, I could get to work on a PR for this, as we do have a need for this in a project that I am currently working on (in order to give us a bit of peace of mind regarding the upcoming phase out 😅). 

Let me know if you have any questions, objections or feedback! Cheers!

EmilePerron:
This PR adds support for partitioned third-party cookies in a backwards compatible way, as described in #1166.

I have updated the documentation's configuration examples, but have not added any tests regarding this new option as the existing tests did not seem to cover similar options. I'll gladly add some if you think it's worth it.

Let me know if you have any comments/feedback!

---

closes #1166

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -110,6 +110,7 @@ class Configuration implements ConfigurationInterface
                             ->scalarNode('domain')->defaultNull()->end()
                             ->scalarNode('secure')->defaultTrue()->end()
                             ->scalarNode('httpOnly')->defaultTrue()->end()
+                            ->scalarNode('partitioned')->defaultFalse()->end()
                             ->arrayNode('split')
                                 ->scalarPrototype()->end()
                             ->end()
--- a/DependencyInjection/LexikJWTAuthenticationExtension.php
+++ b/DependencyInjection/LexikJWTAuthenticationExtension.php
@@ -15,6 +15,7 @@ use Symfony\Component\DependencyInjection\Exception\LogicException;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
+use Symfony\Component\HttpKernel\Kernel;
 
 /**
  * This is the class that loads and manages your bundle configuration.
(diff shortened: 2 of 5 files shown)


EmilePerron:
The CI pipeline errors seem to be unrelated to this PR's changes, and should probably be dealt with in a different PR. I checked other recent PRs and they are all getting this same error, which seems related to an updated dependency.

```diff
--- Expected
+++ Actual
@@ @@
         "title": "LexikJWTAuthenticationBundle",
         "version": "1.0.0"
     },
-    "openapi": "3.0.0",
+    "openapi": "3.1.0",
     "paths": {
         "/login_check": {
             "parameters": [],

/home/runner/work/LexikJWTAuthenticationBundle/LexikJWTAuthenticationBundle/Tests/Functional/Command/ApiPlatformOpenApiExportCommandTest.php:28

FAILURES!
Tests: 166, Assertions: 483, Failures: 1, Skipped: 22.
```

EmilePerron:
Hello! Would it be possible to know when we can expect this to be reviewed and/or merged? 
Or if there is anything prevent this from moving forward :)

maxhelias (approved):

EmilePerron:
Freshly rebased to get the Symfony 7 tests and the OpenAPI 3.1 update that fixes the CI pipeline errors :)

chalasr:
@EmilePerron Awesome. One thing: given the feature can only be used on Symfony 6.4 and higher, should we throw an exception if ones tries to use it with an lower Symfony version? If it makes sense, I'd suggest throwing both from the DI Extension (when setting the option) and the CookieProvider constructor (for potential direct instantiation).

EmilePerron:
> One thing: given the feature can only be used on Symfony 6.4 and higher, should we throw an exception if ones tries to use it with an lower Symfony version? If it makes sense, I'd suggest throwing both from the DI Extension (when setting the option) and the CookieProvider constructor (for potential direct instantiation).

@chalasr Makes sense to me! I just added this in a677b38 :)

chalasr (approved):

chalasr:
Thank you @EmilePerron.

Issue #1137, opened 2023-05-03
Invalidation / Blacklist a token


cedriclombardot:
Hi,

I've not found in documentation, how to invalidate a JWT token, what is the good process method ? 

Found issues about token_blacklist.* but not found in code and not understand the good way 

Thanks in advance

cedriclombardot:
Hi here ? Have u got an idea ? @chalasr is it a missing feature, if true perhaps can i propose an implementation !?

chalasr:
Hi @cedriclombardot, 
Thanks for this and apologies for the late reply, my OSS time is very limited lately.

I want this to happen 👍  So a pull request would be much appreciated. 
Here are some insights about what such an implementation should look like from my POV:
- This feature must be opt-in
- Tokens should be given a  [jti claim](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.7) whose value should be the only thing persisted: if the feature is enabled and a token's jti exists in the blocklist then that token must be rejected.
- Feature detection should not be only based on the presence of the jti, as it mght break existing code that relies on this claim today. 
- The blacklist term should be avoided, alternative such as blocklist should be preferred :)
- We will probably need a simple abstraction for the blocklist storage. A very limited set of built-in implementations should be provided, not necessarily as part of the first iteration (i.e. it can wait til another PR).

Please let me know if you're still wiling to work on this. Anyway, happy to know your thoughts on the above proposal

chalasr:
Implemented in #1170. Thanks for rising the topic @cedriclombardot

PR #1170, opened 2023-11-23
 feat: Invalidate a JWT token

ldaspt:
This PR adds support for invalidating a JWT token #1137.

The code comes mainly from the discussion https://github.com/lexik/LexikJWTAuthenticationBundle/discussions/1005#discussioncomment-2612224  Thanks to @mbabker

I think that the PR meets the needs mentioned in https://github.com/lexik/LexikJWTAuthenticationBundle/issues/1137#issuecomment-1662762926

> 
> * This feature must be opt-in
> * Tokens should be given a jti claim whose value should be the only thing persisted: if the feature is enabled and a token's jti exists in the blocklist then that token must be rejected.
> * Feature detection should not be only based on the presence of the jti, as it mght break existing code that relies on this claim today.
> * The blacklist term should be avoided, alternative such as blocklist should be preferred :)
> * We will probably need a simple abstraction for the blocklist storage. A very limited set of built-in implementations should be provided, not necessarily as part of the first iteration (i.e. it can wait til another PR).
>

### DIFF
--- a/EventListener/AddClaimsToJWTListener.php
+++ b/EventListener/AddClaimsToJWTListener.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Lexik\Bundle\JWTAuthenticationBundle\EventListener;
+
+use Lexik\Bundle\JWTAuthenticationBundle\Event\JWTCreatedEvent;
+
+class AddClaimsToJWTListener
+{
+    public function __invoke(JWTCreatedEvent $event): void
+    {
+        $data = $event->getData();
+
+        if (!isset($data['jti'])) {
+            $data['jti'] = bin2hex(random_bytes(16));
+
+            $event->setData($data);
+        }
+    }
+}
--- a/EventListener/BlockJWTListener.php
+++ b/EventListener/BlockJWTListener.php
@@ -1,0 +1,67 @@
+<?php
+
+namespace Lexik\Bundle\JWTAuthenticationBundle\EventListener;
+
+use Lexik\Bundle\JWTAuthenticationBundle\Exception\JWTDecodeFailureException;
+use Lexik\Bundle\JWTAuthenticationBundle\Exception\MissingClaimException;
+use Lexik\Bundle\JWTAuthenticationBundle\Services\BlockedTokenManagerInterface;
+use Lexik\Bundle\JWTAuthenticationBundle\Services\CacheItemPoolBlockedTokenManager;
+use Lexik\Bundle\JWTAuthenticationBundle\Services\JWTTokenManagerInterface;
+use Lexik\Bundle\JWTAuthenticationBundle\TokenExtractor\TokenExtractorInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Security\Core\Exception\DisabledException;
+use Symfony\Component\Security\Http\Event\LoginFailureEvent;
+use Symfony\Component\Security\Http\Event\LogoutEvent;
+
+class BlockJWTListener
+{
+    private $blockedTokenManager;
+    private $tokenExtractor;
+    private $jwtManager;
+
+    public function __construct(
+        BlockedTokenManagerInterface $blockedTokenManager,
+        TokenExtractorInterface      $tokenExtractor,
+        JWTTokenManagerInterface     $jwtManager
+    ) {
+        $this->blockedTokenManager = $blockedTokenManager;
+        $this->tokenExtractor = $tokenExtractor;
+        $this->jwtManager = $jwtManager;
+    }
+
+    public function onLoginFailure(LoginFailureEvent $event): void
+    {
+        $exception = $event->getException();
+        if (($exception instanceof DisabledException) || ($exception->getPrevious() instanceof DisabledException)) {
+            $this->blockTokenFromRequest($event->getRequest());
+        }
+    }
+
+    public function onLogout(LogoutEvent $event): void
+    {
+        $this->blockTokenFromRequest($event->getRequest());
+    }
+
+    private function blockTokenFromRequest(Request $request): void
+    {
+        $token = $this->tokenExtractor->extract($request);
+
+        if ($token === false) {
+            // There's nothing to block if the token isn't in the request
+            return;
+        }
+
+        try {
+            $payload = $this->jwtManager->parse($token);
+        } catch (JWTDecodeFailureException $e) {
+            // Ignore decode failures, this would mean the token is invalid anyway
+            return;
+        }
+
+        try {
+            $this->blockedTokenManager->add($payload);
+        } catch (MissingClaimException $e) {
+            // We can't block a token missing the claims our system requires, so silently ignore this one
+        }
+    }
+}
--- a/EventListener/RejectBlockedTokenListener.php
+++ b/EventListener/RejectBlockedTokenListener.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Lexik\Bundle\JWTAuthenticationBundle\EventListener;
+
+use Lexik\Bundle\JWTAuthenticationBundle\Event\JWTAuthenticatedEvent;
+use Lexik\Bundle\JWTAuthenticationBundle\Exception\InvalidTokenException;
+use Lexik\Bundle\JWTAuthenticationBundle\Exception\MissingClaimException;
+use Lexik\Bundle\JWTAuthenticationBundle\Services\BlockedTokenManagerInterface;
+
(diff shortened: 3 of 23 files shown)


chalasr:
Nice! Could you please rebase?

ldaspt:
@chalasr Rebase done

EventListener/AddClaimsToJWTListener.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace Lexik\Bundle\JWTAuthenticationBundle\EventListener;
+
+use Lexik\Bundle\JWTAuthenticationBundle\Event\JWTCreatedEvent;
+
+class AddClaimsToJWTListener
+{
+    public function __invoke(JWTCreatedEvent $event): void
```
mbabker, line 9: For a bundle, this is probably better as a named method instead of being invokable.

Also, maybe `Lexik\Bundle\JWTAuthenticationBundle\Services\JWTManager` should be updated to automatically add the `jti` claim to the payload when it creates the token (or at least make it configurable through the bundle config) instead of relying on an event listener.  For the folks who don't want the claim or have their own generation logic, they would only need to disable the feature through the config instead of either having to unload the listener through a compiler pass or mess with listener priorities.

EventListener/RejectBlockedTokenListener.php
```diff
@@ -0,0 +1,29 @@
...
+     */
+    public function __invoke(JWTAuthenticatedEvent $event): void
+    {
+        try {
+            if ($this->tokenManager->has($event->getPayload())) {
+                throw new InvalidTokenException('JWT blocked');
+            }
+        } catch (MissingClaimException) {
+            // Do nothing if the required claims do not exist on the payload (older JWTs won't have the "jti" claim the manager requires)
+        }
```
mbabker: Handling this check inside the authenticators instead of using an event listener might make a little more sense in the context of a bundle (it works well in an app where I wasn't trying to overload the bundle services/behaviors).

Services/BlockedTokenManager.php
```diff
@@ -0,0 +1,74 @@
...
+
+namespace Lexik\Bundle\JWTAuthenticationBundle\Services;
+
+use DateInterval;
+use DateTimeImmutable;
+use DateTimeZone;
+use Lexik\Bundle\JWTAuthenticationBundle\Exception\MissingClaimException;
+use Psr\Cache\CacheItemPoolInterface;
+
+class BlockedTokenManager
```
mbabker: Like I suggested in the discussion, this would probably be good to add an interface for, that way folks can change the storage backend entirely and not rely on a cache store if they want some other method for storing/tracking these.

EventListener/BlockJWTListener.php
```diff
@@ -0,0 +1,59 @@
...
+use Symfony\Component\Security\Http\Event\LogoutEvent;
+
+class BlockJWTListener
+{
+    public function __construct(
+        private BlockedTokenManager $tokenManager,
+        private TokenExtractorInterface $tokenExtractor,
+        private JWTTokenManagerInterface $jwtManager,
+    ) {
+    }
```
mbabker: The bundle still supports PHP 7, so no property promotion.

EventListener/AddClaimsToJWTListener.php
ldaspt: I liked the implementation via a listener because it avoids any modification in the "JWTManager" code and therefore limits regressions

How do you imagine the modification of "JwtManager" ?

I imagine something like that on my side

```
diff --git c/Services/JWTManager.php i/Services/JWTManager.php
index b2008bb..754fc9a 100644
--- c/Services/JWTManager.php
+++ i/Services/JWTManager.php
@@ -44,16 +44,19 @@ class JWTManager implements JWTManagerInterface, JWTTokenManagerInterface
      * @var string
      */
     protected $userIdClaim;
+    private $payloadEnrichments;
 
     /**
-     * @param string|null $userIdClaim
+     * @param JWTEncoderInterface $encoder
+     * @param PayloadEnrichmentInterface[] $payloadEnrichments
      */
-    public function __construct(JWTEncoderInterface $encoder, EventDispatcherInterface $dispatcher, $userIdClaim = null)
+    public function __construct(JWTEncoderInterface $encoder, EventDispatcherInterface $dispatcher, $userIdClaim = null, array $payloadEnrichments = [])
     {
         $this->jwtEncoder = $encoder;
         $this->dispatcher = $dispatcher;
         $this->userIdentityField = 'username';
         $this->userIdClaim = $userIdClaim;
+        $this->payloadEnrichments = $payloadEnrichments;
     }
 
     /**
@@ -64,6 +67,8 @@ class JWTManager implements JWTManagerInterface, JWTTokenManagerInterface
         $payload = ['roles' => $user->getRoles()];
         $this->addUserIdentityToPayload($user, $payload);
 
+        $this->enrichPayload($user, $payload);
+
         return $this->generateJwtStringAndDispatchEvents($user, $payload);
     }
 
@@ -75,9 +80,18 @@ class JWTManager implements JWTManagerInterface, JWTTokenManagerInterface
         $payload = array_merge(['roles' => $user->getRoles()], $payload);
         $this->addUserIdentityToPayload($user, $payload);
 
+        $this->enrichPayload($user, $payload);
+
         return $this->generateJwtStringAndDispatchEvents($user, $payload);
     }
 
+    private function enrichPayload(UserInterface $user, array &$payload): void
+    {
+        foreach ($this->payloadEnrichments as $payloadEnrichment) {
+            $payloadEnrichment->enrich($user, $payload);
+        }
+    }
+
     /**
      * @return string The JWT token
      */
diff --git c/Services/PayloadEnrichment/RandomJtiEnrichment.php i/Services/PayloadEnrichment/RandomJtiEnrichment.php
new file mode 100644
index 0000000..82a8926
--- /dev/null
+++ i/Services/PayloadEnrichment/RandomJtiEnrichment.php
@@ -0,0 +1,14 @@
+<?php
+
+namespace Lexik\Bundle\JWTAuthenticationBundle\Services\PayloadEnrichment;
+
+use Lexik\Bundle\JWTAuthenticationBundle\Services\PayloadEnrichmentInterface;
+use Symfony\Component\Security\Core\User\UserInterface;
+
+class RandomJtiEnrichment implements PayloadEnrichmentInterface
+{
+    public function enrich(UserInterface $user, array &$payload): void
+    {
+        $payload['jti'] = bin2hex(random_bytes(16));
+    }
+}
diff --git c/Services/PayloadEnrichmentInterface.php i/Services/PayloadEnrichmentInterface.php
new file mode 100644
index 0000000..5278623
--- /dev/null
+++ i/Services/PayloadEnrichmentInterface.php
@@ -0,0 +1,10 @@
+<?php
+
+namespace Lexik\Bundle\JWTAuthenticationBundle\Services;
+
+use Symfony\Component\Security\Core\User\UserInterface;
+
+interface PayloadEnrichmentInterface
+{
+    public function enrich(UserInterface $user, array &$payload): void;
+}
```

EventListener/BlockJWTListener.php
ldaspt: fixed

Services/BlockedTokenManager.php
ldaspt: `BlockedTokenManagerInterface` interface and a cache implementation added  `CacheItemPoolBlockedTokenManager`

EventListener/AddClaimsToJWTListener.php
mbabker: I like that enrichments idea.

For me, the difference between having the `JWTManager` add the `jti` claim versus an event listener is that I feel like the manager should build the full payload that it is configured to create and then event listeners are the third-party hook to manipulate that payload.  Neither approach is wrong, it just kind of boils down to what folks decide is the approach the core library's going to take.

Vincz:
Hi guys! Any ETA regarding this feature? Thanks!

chalasr:
Hey @Vincz :) I'll do my best to move forward with this in the 2 coming weeks, expect it to be released in March

ldaspt:
@chalasr Rebase done, whenever you have some free time, could you please take a look at the pull request? I would really appreciate your feedback and suggestions. Let me know if there's anything specific you'd like me to address.

Thanks!

chalasr:
Thank you @ldaspt.

chalasr:
Follow-up PR to give a try to [@mbabker's suggestion](https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1170#discussion_r1406341291) welcome!

PR #1176, opened 2023-12-03
Remove ECS in favor of PHPCSFixer

chalasr:


.github/workflows/coding-standards.yml
```diff
@@ -8,12 +8,12 @@ jobs:
...
             matrix:
                 operating-system: [ubuntu-latest]
-                php-versions: ['8.1']
+                php-versions: ['8.2']
         name: PHP ${{ matrix.php-versions }} Test on ${{ matrix.operating-system }}
 
         steps:
             - name: Checkout
-              uses: actions/checkout@v3
+              uses: actions/checkout@v4
```
theofidry, line 16: this could probably be extracted

.github/workflows/coding-standards.yml
```diff
@@ -8,12 +8,12 @@ jobs:
         strategy:
             matrix:
                 operating-system: [ubuntu-latest]
-                php-versions: ['8.1']
+                php-versions: ['8.2']
```
theofidry, line 11: this could probably be extracted

.github/workflows/coding-standards.yml
```diff
@@ -27,12 +27,14 @@ jobs:
             - name: Install Composer dependencies
               run: |
                   composer install
-                  composer require rector/rector symplify/easy-coding-standard --dev
+                  mkdir -p tools/php-cs-fixer tools/rector
+                  composer require --working-dir=tools/php-cs-fixer friendsofphp/php-cs-fixer --dev
+                  composer require rector/rector --dev
```
theofidry, line 32: For this use case I would recommend using [composer-bin-plugin](https://github.com/bamarni/composer-bin-plugin) with ramsey/install:

```yaml
            -   name: Install PHP-CS-Fixer Composer bin dependencies
                uses: ramsey/composer-install@v2
                with:
                    working-directory: vendor-bin/php-cs-fixer
```

This would have three advantages:

- You can easily install the tool locally to run it locally too.
- The installation of the deps does not require any dependency resolution on Composer's end and the download will be cached.
- Since those are locked dependencies, you can use dependabot to periodically upgrade them:

```yaml
    -   package-ecosystem: "composer"
        directory: "vendor-bin/*/"
        schedule:
            interval: "weekly"
        groups:
            dependencies:
                patterns:
                    - "*"
```

I would say though for rector I put a `*` as a dependency to upgrade whenever possible.

Erratum as I see you committed `tools/php-cs-fixer/composer.json`:

- You can configure bamarni-composer-bin-plugin to rename `vendor-bin` into `tools`: https://github.com/bamarni/composer-bin-plugin#target-directory-target-directory
- I recommend you to lock the dependencies, i.e. publish the `composer.lock` for rector and PHP-CS-Fixer.
- You probably don't want to do a `composer require` in the case above

.github/workflows/coding-standards.yml
```diff
@@ -27,12 +27,14 @@ jobs:
...
 
-            - name: CONDING STANDARDS (ECS)
+            - name: CODING STANDARDS (ECS)
```
theofidry:
```suggestion
            - name: PHP-CS-Fixer
```
? (if you have another cs fixer, e.g. for YAML files, you'll probably put it in a dedicated step

.github/workflows/coding-standards.yml
```diff
@@ -27,12 +27,14 @@ jobs:
...
               run: |
-                  vendor/bin/ecs check
+                  tools/php-cs-fixer/vendor/bin/php-cs-fixer fix .
 
             - name: CONDING STANDARDS (RECTOR)
               run: |
-                  vendor/bin/rector process --ansi --dry-run --xdebug
+                  tools/rector/vendor/bin/rector process . --ansi --dry-run --xdebug
```
theofidry, line 40: why is it running with Xdebgu?

.php-cs-fixer.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+$finder = PhpCsFixer\Finder::create()->in([__DIR__]);
+
+return (new PhpCsFixer\Config())
```
theofidry, line 5: if you feel like it: https://github.com/theofidry/php-cs-fixer-config

Or alternatively create your own, it's relatively easy.

.php-cs-fixer.php
```diff
@@ -0,0 +1,14 @@
...
+    ->setRules([
+        '@Symfony' => true,
+        'ordered_imports' => true,
+        'array_syntax' => ['syntax' => 'short'],
+        'phpdoc_scalar' => false,
+    ])
+    ->setUsingCache(false)
```
theofidry, line 12: no cache? Personally I like to use it but I put the output in a `dist/` directory

.github/workflows/coding-standards.yml
chalasr: done in #1179

.github/workflows/coding-standards.yml
chalasr: done in #1178

PR #1178, opened 2023-12-04
[GHA] Infra & build updates

chalasr:


.github/workflows/web-token-ci.yml
```diff
@@ -63,7 +64,7 @@ jobs:
                 composer require --dev --no-update web-token/jwt-encryption-algorithm-aesgcmkw:"^3.0.6"
 
             - name: "Install dependencies"
-              run: "composer update ${{ matrix.composer-flags }} --prefer-dist"
+              run: "composer update ${{ matrix.composer-flags }} --prefer-dist --prefer-stable"
```
theofidry, line 67: I would recommend to use `ramsey/composer-install` for caching. It also removes the need of `--prefer-dist --prefer-stable` and then doing `--prefer-lowest` or not is defined by `with-dependencies: highest|lowest`

theofidry: You might need to put `SYMFONY_REQUIRE` as an environment variable though, but you can do so like that:

```yaml
            -   name: Configure the Symfony version
                run: echo "SYMFONY_REQUIRE=${{ matrix.symfony }}" >> $GITHUB_ENV
```

PR #1180, opened 2023-12-05
Allow setting advanced cookie name by dedicated configuration attribute

webda2l:
Allow configurations like:

```
lexik_jwt_authentication:
    ...
    token_extractors:
        split_cookie:
            enabled: true
            cookies:
                - jwt_hp
                - jwt_s
    set_cookies:
        jwt_hp:
            customName: '%env(JWT_COOKIE_PREFIX)%jwt_hp'
            ...
        jwt_s:
            customName: '%env(JWT_COOKIE_PREFIX)%jwt_s'
            ...
```

With Gesdinet, it's already possible with:

```
gesdinet_jwt_refresh_token:
    token_parameter_name: '%env(JWT_COOKIE_PREFIX)%refresh_token'
    ...
```

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -98,6 +98,7 @@ class Configuration implements ConfigurationInterface
                     ->useAttributeAsKey('name')
                     ->prototype('array')
                         ->children()
+                            ->scalarNode('name')->defaultNull()->end()
                             ->scalarNode('lifetime')
                                 ->defaultNull()
                                 ->info('The cookie lifetime. If null, the "token_ttl" option value will be used')
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -98,7 +98,7 @@ class Configuration implements ConfigurationInterface
                     ->useAttributeAsKey('name')
                     ->prototype('array')
                         ->children()
-                            ->scalarNode('name')->defaultNull()->end()
+                            ->scalarNode('customName')->defaultNull()->end()
                             ->scalarNode('lifetime')
                                 ->defaultNull()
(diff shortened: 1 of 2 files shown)


maxhelias:
Hummm.. It's already possible, the attributes key is the name. ex : `jwt_hp` & `jwt_s`

webda2l:
```
lexik_jwt_authentication:
    ...
    token_extractors:
        split_cookie:
            enabled: true
            cookies:
                - '%env(JWT_COOKIE_PREFIX)%jwt_hp'
                - '%env(JWT_COOKIE_PREFIX)%jwt_s'
    set_cookies:
        '%env(JWT_COOKIE_PREFIX)%jwt_hp':
            lifetime: 2592000
           ...
        '%env(JWT_COOKIE_PREFIX)%jwt_s':
            lifetime: 2592000
           ...
```

works but IMO, when the cookie naming require a more advanced syntax like in my case, it's less conventional and clean than using a dedicated attribute.
My proposition is optional for ease some advanced cases only.

And the current and basic attribute key for basic naming remains useful for service naming  https://github.com/lexik/LexikJWTAuthenticationBundle/blob/2.x/DependencyInjection/LexikJWTAuthenticationExtension.php#L125 and in the token_extractors.[cookie|split_cookie] part.

maxhelias:
For advanced syntax, I'd recommend using the parameters instead.

In your case : you can setting a `customName` and attributes key which have 2 different behaviors. The token extractors doesn't use the customName but the attributes key and the cookie name isn't the attributes key, it's a mix and imo these changes bring more complexity to the bundle.

chalasr:
I agree with @maxhelias this looks rather confusing. Thanks for the PR, I hope to see you again for another.

PR #1184, opened 2023-12-07
Wrong type exception for AccessTokenLoader with certain configuration

athanasius-kircher:
If I use the feature "webtoken" with that bundle than the following configuration (only signing jwk but not encrypting), leads to an exception when the Encoder is loader and Dependencies are injected.

```
lexik_jwt_authentication:
    public_key: '%env(resolve:JWT_PUBLIC_KEY)%'
    encoder:
        service: lexik_jwt_authentication.encoder.web_token
    access_token_issuance:
        enabled: true
        signature:
            algorithm: EdDSA
            key: '%env(SIGNATURE_KEY)%'
    access_token_verification:
        enabled: true
        signature:
            allowed_algorithms:
                - EdDSA
            keyset: '%env(SIGNATURE_VERIFICATION_KEYSET)%'
```
The problem in my opinion is the default xml Service configuration, that does not specify the right type. If one acitvates encryption this error disappears as than this config is overwritten anyhow.


```
In AccessTokenLoader.php line 38:                                                                                                                                                                                                                                                                        
Lexik\Bundle\JWTAuthenticationBundle\Services\WebToken\AccessTokenLoader::__construct(): Argument #10 ($jweHeaderChecker) must be of type ?array, string given, called in /var/www/html/var/cache/test/Container1K8ziZx/getLexikJwtAuthentication_AccessTokenLoaderService.php on line 29 
```

### DIFF
--- a/Resources/config/web_token_verification.xml
+++ b/Resources/config/web_token_verification.xml
@@ -9,15 +9,15 @@
             <argument type="service" id="Jose\Bundle\JoseFramework\Services\JWSLoaderFactory" />
             <argument type="service" id="Jose\Bundle\JoseFramework\Services\JWELoaderFactory" on-invalid="null" />
             <argument type="service" id="Jose\Bundle\JoseFramework\Services\ClaimCheckerManagerFactory" />
-            <argument /> <!-- Claim checkers -->
-            <argument /> <!-- JWS header checkers -->
-            <argument /> <!-- Mandatory claims -->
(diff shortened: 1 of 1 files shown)


chalasr:
Thanks for the PR! We miss some tests apparently... I'll take care of that.  I'm gonna merge+tag this PR asap by the end of the week.

chalasr:
Thank you @athanasius-kircher.

PR #1187, opened 2023-12-23
feat: Pass DisabledException to JWTAuthenticationFailureResponse

arneee:
Hi,

currently the response is always "Bad credentials". If a user has been disabled, the response should be accordingly so the user can take additional steps, for example contacting the admin.

This PR checks if the AuthenticationException "previous" is a DisabledException and passes that instead.

What do you think?


mbabker:
This seems like a very application-specific thing to me.  Passing through whatever exception the authentication system gives to the failure handler is the best thing for the bundle to do IMO, and an application can make checks like this to adjust the response either in a decorated failure handler or through the authentication failure event that is dispatched.

PR #1183, opened 2023-12-06
Fix for PHP <7.4 compatibility

fracsi:
Follow up for #1174
Fixes #1182 
Arrow functions are introduced in php 7.4 as well.

PR #1188, opened 2023-12-24
[CI] Fixed CI for PHP 7.2

mnocon:
This PR is my attempt at fixing the "always green" build for 7.2 (mentioned in https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1183#issuecomment-1843291457)

Things done:
1) I've bumped the PHPUnit Bridge version to the latest one.

According to [PHPUnit Bridge doc](https://symfony.com/doc/current/components/phpunit_bridge.html#installation):
```
The PHPUnit bridge is designed to work with all maintained versions of Symfony components, even across different major versions of them. You should always use its very latest stable major version to get the most accurate deprecation report.
```

Allowing phpunit-bridge range of `"^4.4|^5.4|^6.0|^7.0"` together with `--prefer-lowest` means that the 4.4.0 version of it is installed - which doesn't have all the required bugfixes.

I've chosen ^7.0 as the latest one, for the 7.x series 7.0.1 is needed because it contains the bugfix for this issue: https://github.com/symfony/demo/pull/1457#issuecomment-1836188457

2) Set PHPUnit 9.5 as the max version

The 7.2 build cannot use PHPUnit 9.5, [it requires PHP 7.3](https://github.com/sebastianbergmann/phpunit/blob/9.5.0/composer.json#L24)

I've set this as max version instead - it will be used if possible, otherwise lower versions will be used.

3) Bumped lowest supported version of lcobucci/jwt to 3.4.6

One of the tests was failing:
```
1) Lexik\Bundle\JWTAuthenticationBundle\Tests\Functional\GetTokenTest::testGetTokenWithCustomClaim
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
 Array &0 (
     0 => 'foo'
-    1 => 'bar'
 )
```

This was fixed by bumping the version of lcobucci/jwt to 3.4.6 - I suspect it contains bugfix not present in 3.4.0

4) Run PHP 7.2 tests on ubuntu-20

I found this issue when googling: https://github.com/shivammathur/setup-php/issues/720
```
If you are running this on the ubuntu-latest or ubuntu-22.04 runner, that has OpenSSL 3 and support for it was added to PHP in 8.1. Please use ubuntu-20.04 runner with the older PHP versions and this should work.
```

And it was right, all of a sudden the segmentation faults have disappeared.

Overall I feel everything I did was a bit hacky, but the build is now green 🎉

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
     tests:
-        runs-on: ubuntu-latest
+        runs-on: ${{ matrix.os }}
         continue-on-error: ${{ matrix.can-fail }}
         strategy:
             fail-fast: false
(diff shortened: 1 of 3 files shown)


chalasr:
Thanks a lot @mnocon, much appreciated!

PR #1204, opened 2024-02-10
Use symfony/clock component when available instead of non-testable time()

Dean151:
**Motivation**:
Tests are already hard enough to build so that they are testing, covering, isolated and in an expected way.
Therefore, managing external dependencies is an obligation. And time is an external dependency.

With the introduction of [symfony/clock](https://symfony.com/components/Clock) in 6.3, Symfony proposes a solution for managing time in an application, with a shared instance that is overridable during the tests.

Now that the library requires it as a dependencies, we can use it everywhere it should be used in place of php native methods.

**Solution**:
This PR looked over the time() usage, and replaced them with Symfony's `Clock::now()->getTimeStamp()` when available.

**Retro-compatibility**:
This behavior change is probably what the user would expect from a library having symfony/clock as a requirement ; so they probably are fine.

**Tests**:
Current test suite still uses time(), but they could be improved to take advantage of this new testability.

### DIFF
--- a/Command/GenerateTokenCommand.php
+++ b/Command/GenerateTokenCommand.php
@@ -3,6 +3,7 @@
 namespace Lexik\Bundle\JWTAuthenticationBundle\Command;
 
 use Lexik\Bundle\JWTAuthenticationBundle\Services\JWTTokenManagerInterface;
+use Symfony\Component\Clock\Clock;
 use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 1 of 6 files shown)


Dean151:
Sadly, this might be incompatible with:
- https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1155
- https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1157

Therefore, I do think the dependency should be flexible in the source, and we could use the one we prefer in the tests so that the end user can rely on whatever psr/clock implementation he prefers

chalasr:
Apologies for being one year late here. I would like to replace lcobucci/clock by symfony/clock and have symfony/clock as dev-dependency. The former requires to add some backward compatibility layers. Once done I'll merge your PR (without the class-exists checks)

Dean151:
> Apologies for being one year late here. I would like to replace lcobucci/clock by symfony/clock and have symfony/clock as dev-dependency. The former requires to add some backward compatibility layers. Once done I'll merge your PR (without the class-exists checks)

No need to apologize! Open-source isn’t a full-time job, and I truly appreciate the time you spend maintaining, testing, and reviewing. Thank you! 😊

Dean151:
Hello @chalasr !
In your own time, of course, just be aware that I rebased my PR to target 3.x branch, and took advantage of the removal of `lcobucci/clock` to simplify the code change.

With my best sentiments!

chalasr:
It took time, but here we go, this is in now. Thank you very much @Dean151.

PR #1208, opened 2024-02-23
Remove security for `/auth` route in OpenAPI generator

Arkanii:
The `/auth` route is always public, so we can disabled the security part in OpenAPI.


Arkanii:
I don't understand the `CI Tests` error, I don't think this is related to the PR.

Issue #1209, opened 2024-02-27
Wrong definition of AccessTokenBuilder


CvekCoding:
Hello.

I have encountered a problem, and it seems that it is related to an incorrect definition in the file `Resources/config/web_token_issuance.xml `:
```
        <service id="lexik_jwt_authentication.access_token_builder" class="Lexik\Bundle\JWTAuthenticationBundle\Services\WebToken\AccessTokenBuilder" public="false">
            <argument type="service" id="Symfony\Contracts\EventDispatcher\EventDispatcherInterface" />
            <argument type="service" id="Jose\Bundle\JoseFramework\Services\JWSBuilderFactory" />
            <argument type="service" id="Jose\Bundle\JoseFramework\Services\JWEBuilderFactory" on-invalid="null" />
            <argument /> <!-- Signature algorithm -->
            <argument /> <!-- Signature key -->
            <argument on-invalid="null" /> <!-- Key encryption algorithm -->
            <argument on-invalid="null" /> <!-- Content encryption algorithm -->
            <argument on-invalid="null" /> <!-- Encryption key -->
        </service>
```
You have the 5th, 6th and 7th arguments, which must be nulls in case we don't use encryption. But it doesn't seem that Symfony works that way - for text arguments, `on-invalid="null"` directive just doesn't work, the value of these arguments is an empty string ''. So the service `Services/WebToken/AccessTokenBuilder.php ` receives these arguments and does not work correctly.

As a solution, I would suggest in the file `DependencyInjection/LexikJWTAuthenticationExtension.php ` explicitly prescribe values in these arguments for the case when we do not encrypt the keys. Like this:
```
            if ($config['access_token_issuance']['encryption']['enabled'] === true) {
                $accessTokenBuilderDefinition
                    ->replaceArgument(5, $config['access_token_issuance']['encryption']['key_encryption_algorithm'])
                    ->replaceArgument(6, $config['access_token_issuance']['encryption']['content_encryption_algorithm'])
                    ->replaceArgument(7, $config['access_token_issuance']['encryption']['key'])
                ;
            } else {
                $accessTokenBuilderDefinition
                    ->replaceArgument(5, null)
                    ->replaceArgument(6, null)
                    ->replaceArgument(7, null);
            }
```

[...]

PR #1211, opened 2024-02-28
Fix encryption

CvekCoding:
For explanation see https://github.com/lexik/LexikJWTAuthenticationBundle/issues/1209

### DIFF
--- a/Resources/config/web_token_issuance.xml
+++ b/Resources/config/web_token_issuance.xml
@@ -11,9 +11,9 @@
             <argument type="service" id="Jose\Bundle\JoseFramework\Services\JWEBuilderFactory" on-invalid="null" />
             <argument /> <!-- Signature algorithm -->
             <argument /> <!-- Signature key -->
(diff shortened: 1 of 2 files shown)


chalasr:
Thanks for the great report and the PR. I think this has been fixed in #1214 (+ patch on web-token's library) so I'm closing.

SCIF:
Hi @chalasr , 

But the problem is still there in 2.x branch so could you please consider this issue as relevant? May be backport a proper solution required. I have a fresh install of symfony 7.x with this bundle of version 2.20.3. Using of web-token is breaking because of this issue

chalasr:
👋 @SCIF 
The fix has been [backported](https://github.com/lexik/LexikJWTAuthenticationBundle/commit/60770f1fb9ec0b8b0f27a0ee9e2bf441ca4a1db9) but not tagged yet! Expect 2.20.4 to be released tomorrow

SCIF:
Hi @chalasr ,

Just a courtesy reminder :)

PR #1214, opened 2024-03-23
Fix ci 3x

Spomky:
Base on #1201 
/cc @chalasr

PR #1225, opened 2024-06-06
Fix default values in WebToken services when encryption disabled

### LINKED ISSUE #1223: The algorithm with the alias "" is not supported
| Q | A
| --- | ---
| Symfony Version | 6.4.7
| Bundle Version | 3.0.0
| PHP Version | 8.3.7
| Related issues/PRs | #1209, #1214 

I've been attempting to configure this bundle to use the Web-Token feature as [outlined in the documentation](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/b20c4ae7fdfe1d7422c7099a141cc9eb64627310/Resources/doc/10-web-token.rst) and have run into some issues.

I have the following configuration:
```yaml
lexik_jwt_authentication:
  secret_key: '%env(resolve:JWT_SECRET_KEY)%'
  public_key: '%env(resolve:JWT_PUBLIC_KEY)%'
  pass_phrase: '%env(JWT_PASSPHRASE)%'
  api_platform:
    enabled: true
    check_path: /api/login_check
    username_path: username
    password_path: password
  encoder:
    service: lexik_jwt_authentication.encoder.web_token
  token_ttl: 3600
  allow_no_expiration: false
  clock_skew: 0
  user_id_claim: username
  token_extractors:
    authorization_header:
      enabled: true
      prefix: Bearer
      name: Authorization
    cookie:
      enabled: false
      name: BEARER
    query_parameter:
      enabled: false
      name: bearer
    split_cookie:
      enabled: false
      cookies: {  }
  remove_token_from_body_when_cookies_used: true
  set_cookies: {  }
  access_token_issuance:
    enabled: true
    signature:
      algorithm: RS256
      key: # some key
  access_token_verification:
    enabled: true
    signature:
      allowed_algorithms:
        - RS256
      keyset: # keyset
  blocklist_token:
    enabled: false
    cache: cache.app
```

If encryption isn't enabled, an `InvalidArgumentException` is always thrown by `AccessTokenBuilder` whenever a request for a token is made. For example, the following always fails when encryption isn't enabled:

```shell
curl -X POST http://localhost/api/login_check \
-H "Content-Type: application/json" \
-d '{"username":"someusername","password":"somepassword"}'
```

The problematic lines of code appear to be here: https://github.com/lexik/LexikJWTAuthenticationBundle/blob/b20c4ae7fdfe1d7422c7099a141cc9eb64627310/Services/WebToken/AccessTokenBuilder.php#L68-L71

`$keyEncryptionAlgorithm` and `$contentEncryptionAlgorithm` aren't automatically set to `null` if they aren't defined and instead are set to empty strings. This always causes an exception to the thrown due to the encryption algorithm being set to an empty string. If I manually set both variables to `null` before the above lines of code, my configuration works and a token is correctly sent in the response.

The encryption related service arguments are replaced here, but only if encryption is enabled:

https://github.com/lexik/LexikJWTAuthenticationBundle/blob/b20c4ae7fdfe1d7422c7099a141cc9eb64627310/DependencyInjection/LexikJWTAuthenticationExtension.php#L217-L224

Otherwise, they remain as empty strings, which doesn't appear to be the intention since it makes encryption mandatory.

I've attempted to directly set the encryption configuration options to `null` as a possible workaround, but that causes a different exception to be thrown that says they can't be `null`. That seems to directly contradict the service responsible for building tokens.

Am I missing an important configuration option/detail that addresses these issues or is this a bug? I've gone through the code and documentation in an attempt to find such an option and haven't had any luck. Is encryption supposed to be required when using the web-token feature?

NeuralClone:
| Q | A
| --- | ---
| Symfony Version | 6.4.x
| Bundle Version | 3.0.0
| PHP Version | 8.2
| Fixed Tickets | fixes #1223 
| Related Issues/PRs | #1209, #1214 

This PR fixes the default values in the `AccessTokenBuilder` and `AccessTokenLoader` services when encryption is disabled.

### DIFF
--- a/Resources/config/web_token_issuance.xml
+++ b/Resources/config/web_token_issuance.xml
@@ -11,9 +11,9 @@
             <argument type="service" id="Jose\Bundle\JoseFramework\Services\JWEBuilderFactory" on-invalid="null" />
             <argument /> <!-- Signature algorithm -->
             <argument /> <!-- Signature key -->
-            <argument /> <!-- Key encryption algorithm -->
-            <argument /> <!-- Content encryption algorithm -->
-            <argument /> <!-- Encryption key -->
+            <argument>null</argument> <!-- Key encryption algorithm -->
+            <argument>null</argument> <!-- Content encryption algorithm -->
+            <argument>null</argument> <!-- Encryption key -->
         </service>
     </services>
 </container>
--- a/Resources/config/web_token_verification.xml
+++ b/Resources/config/web_token_verification.xml
@@ -10,15 +10,15 @@
             <argument type="service" id="Jose\Bundle\JoseFramework\Services\JWELoaderFactory" on-invalid="null" />
             <argument type="service" id="Jose\Bundle\JoseFramework\Services\ClaimCheckerManagerFactory" />
             <argument type="collection" /> <!-- Claim checkers -->
-            <argument type="collection"/> <!-- JWS header checkers -->
-            <argument type="collection"/> <!-- Mandatory claims -->
+            <argument type="collection" /> <!-- JWS header checkers -->
+            <argument type="collection" /> <!-- Mandatory claims -->
             <argument type="collection" /> <!-- Allowed signature algorithms -->
             <argument /> <!-- Signature keyset -->
-            <argument on-invalid="null" /> <!-- Continue on decryption failure -->
+            <argument>false</argument> <!-- Continue on decryption failure -->
             <argument type="collection" /> <!-- JWE header checkers -->
             <argument type="collection" /> <!-- Allowed key encryption algorithms -->
             <argument type="collection" /> <!-- Allowed content encryption algorithms -->
-            <argument on-invalid="null" /> <!-- Encryption keyset -->
+            <argument>null</argument> <!-- Encryption keyset -->
         </service>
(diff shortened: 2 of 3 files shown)


SCIF:
Any chance to review this? This affects fresh installations using web-token.

NeuralClone:
A review for this PR would definitely be greatly appreciated.

In the original issue I opened, I posted a follow-up comment with a workaround that involves adding a compiler pass to your project. Here's a direct link to it in case you missed it:

https://github.com/lexik/LexikJWTAuthenticationBundle/issues/1223#issuecomment-2146627199

It doesn't actually solve the configuration problem, of course. It also assumes that encryption will remain disabled.

But it should let you use the web token services until this PR is reviewed/merged or another solution is implemented.

chalasr (approved):

maxhelias (approved):

chalasr:
Thank you @NeuralClone.

SCIF:
Hi @chalasr , any chance to compose a release in order to make these changes available?

chalasr:
Released in https://github.com/lexik/LexikJWTAuthenticationBundle/releases/tag/v3.1.1

Issue #1224, opened 2024-05-31
deprecation


lxregistry:
The "Symfony\Component\HttpKernel\DependencyInjection\Extension" class is considered internal since Symfony 7.1, to be deprecated in 8.1; use Symfony\Component\DependencyInjection\Extension\Extension instead. It may change without further notice. You should not use it from "Lexik\Bundle\JWTAuthenticationBundle\DependencyInjection\LexikJWTAuthenticationExtension".

Chris53897:
@lxregistry Can be closed as completed. Please see linked PR

PR #1226, opened 2024-06-10
feature: fix deprecation of Extension

Chris53897:
`The "Symfony\Component\HttpKernel\DependencyInjection\Extension" class is considered internal since Symfony 7.1, to be deprecated in 8.1; use Symfony\Component\DependencyInjection\Extension\Extension instead. It may change without further notice. You should not use it from "Lexik\Bundle\JWTAuthenticationBundle\DependencyInjection\LexikJWTAuthenticationExtension".`


maxhelias (approved):

Chris53897:
Will close https://github.com/lexik/LexikJWTAuthenticationBundle/issues/1224

theofidry (approved):
I would not call this a feature but 👍

stof (approved):

chalasr:
Thank you @Chris53897.

PR #1235, opened 2024-07-31
Update 10-web-token.rst

cosminsandu:
Update documentation headers (from markdown style) to restructuredtext.

This commit fixes the headers in the official documentation from: https://symfony.com/bundles/LexikJWTAuthenticationBundle/current/10-web-token.html


chalasr:
Good catch, thanks @cosminsandu.

Issue #1040, opened 2022-06-16
JWTManager uses undefined function "getCredentials" of class "Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken"


nourheine:
Hi guys!
I'm using symfony 6 (after i upgrade from version 5.3) , I installed the bundle lexikJWTAuthenticationBundle, now i intent to decode the created token so i refere to this documentation: 
https://symfony.com/bundles/LexikJWTAuthenticationBundle/current/9-access-authenticated-jwt-token.html

I got this error when i use this code to decode the generated token 
Attempted to call an undefined method named "getCredentials" of class "Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken"

I suppose that the decode function in Lexik\Bundle\JWTAuthenticationBundle\Services\JWTManager uses the function getCredentials that is deprecated in symfony 5.4 and removed in symfony 6.

I will appreciate any help! thanks in advance.

fmarchalemisys:
Hi,

I just got hit by this bug.

The failure is caused by `\Lexik\Bundle\JWTAuthenticationBundle\Services\JWTManager::decode()` declared as expecting a `TokenInterface` which, as you pointed out, has no `getCredentials()` anymore.

It works when called with a `JWTUserToken`, `PreAuthenticationJWTUserToken` or `JWTPostAuthenticationToken` because they all define the `getCredentials()` method.

Unfortunately, no interface is defined to easily check the provided token is compatible with `JWTManager::decode()`.

The only solution on your side is to check the method exists on the token before trying to decode it:

```php
if (!\is_object($token) || !method_exists($token, "getCredentials")) {
    throw new \Exception("Not a jwt token");
}
$payload = $jwtManager->decode($token);
if ($payload === false) {
    throw new \Exception("Invalid jwt token");
}
```

Are you guys open to a PR adding a `JWTTokenInterface` and updating `JWTManager::decode(JWTTokenInterface $token)` (it will break custom tokens not implementing the new interface) or is the problem more structural than this (i.e. must get rid of `getCredentials()` altogether) or should the above check be included in `JWTManager::decode()` to cleanly return `false` while still accepting any `TokenInterface` ?

chalasr:
Hi, sorry about the super late reply. but I'd need some reproducing code in order to understand the issue and eventually fix it. Please comment if you can provide that.

fmarchalemisys:
Hi @chalasr ,

[...]

PR #1244, opened 2024-11-02
Ignore token not having the getCredentials() method during decode

fmarchalemisys:
The `getCredentials()` method  was removed from `TokenInterface` in Symfony 6. Beside, some tokens, such as `TestBrowserToken`, return null instead of the expected string type. See [issue #1040](https://github.com/lexik/LexikJWTAuthenticationBundle/issues/1040).

Decoding a jwt token is only meaningful when the token is a `JWTPostAuthenticationToken`. Any other token can be ignored.

To fix the deprecation, the user's code must do something like this:

```php
if (!$token instanceof JWTPostAuthenticationToken) {
    return;
}
try {
    $jwtPayload = $this->jwtManager->decode($token);
} catch (JWTDecodeFailureException) {
    return;
}
if ($jwtPayload === false) {
    return;
}
// process $jwtPayload…    
```

Now, the questions:

1. Do we deprecate the call with anything other than a `JWTPostAuthenticationToken` with the aim of restricting the type to `public function decode(JWTPostAuthenticationToken $token): array|bool` later on?
2. Do we simply ignore tokens other than `JWTPostAuthenticationToken` without deprecation to make the use of the decode method easier?

This PR is currently aiming at the first solution but solution 2 simply requires to remove the `trigger_deprecation` call.

### DIFF
--- a/Services/JWTManager.php
+++ b/Services/JWTManager.php
@@ -10,6 +10,7 @@ use Lexik\Bundle\JWTAuthenticationBundle\Event\JWTEncodedEvent;
 use Lexik\Bundle\JWTAuthenticationBundle\Events;
 use Lexik\Bundle\JWTAuthenticationBundle\Exception\JWTDecodeFailureException;
 use Lexik\Bundle\JWTAuthenticationBundle\Exception\JWTEncodeFailureException;
+use Lexik\Bundle\JWTAuthenticationBundle\Security\Authenticator\Token\JWTPostAuthenticationToken;
 use Lexik\Bundle\JWTAuthenticationBundle\Services\PayloadEnrichment\NullEnrichment;
 use Symfony\Component\PropertyAccess\PropertyAccess;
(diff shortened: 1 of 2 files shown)


chalasr:
Thanks for pushing the topic and doing the PR, that's much appreciated.
However, I'm not sure ignoring any other token type is the right approach as we'll have to stop implementing `getCredentials()` in `JWTPostAuthenticationToken` anyway since it is deprecated.
Here is a PR for 2.x (which is still maintained for bugfixes) that makes use of `TokenInterface::get/setAttribute()` instead: https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1251.
For 3.x, I'm thinking about deprecating `JWTManager::decode()` entirely, suggesting `parse()` as replacement. Both `JWTPostAuthenticationToken` + `decode(TokenInterface)` feels overcomplicated. Any thought on this is welcome

janedbal:
Any plans here? We just got hit by that 3 years old bug (#1040), but it is still not resolved.

Issue #1258, opened 2024-11-29
JWTCookieProvider::createCookie does not set correct values when some params are provided false


nuryagdym:
**$secure, $httpOnly, $partitioned** params of `JWTCookieProvider::createCookie` are not used when they are `false` because of condition used.

issue exist on both 2.x and 3.x version.

Opened MR for 3.x branch, can also open another mr for 2.x if neccessary.

PR #1259, opened 2024-11-29
fix issue #1258 JWTCookieProvider does set flags cookie flags when va…

nuryagdym:
fixes issue #1258


maxhelias (approved):

chalasr:
Thank you @nuryagdym.

PR #1267, opened 2025-01-06
Fix wrongly silenced signing failure

chalasr:
Thanks @backendtea for the patch!


BackEndTea:
Thanks for the merge!

PR #1276, opened 2025-06-04
Update LcobucciJWSProvider.php

Bartheyrman22:
Would be great to add some flexibility to this class to extend it. Otherwise we are more or less forced to copy it completely when needing to change the TTL on a user basis for instance.


VincentLanglet:
What's the use case ? You can still use composition over extension.

maxhelias:
This class is marked as `@final` so it's a bad idea to extend it.

PR #1277, opened 2025-06-04
Improve JWTUser phpdoc

VincentLanglet:
Hi @chalasr,

Since https://github.com/symfony/symfony/pull/58007 the `getUserIdentifier` method is supposed to return a non-empty-string, which means that
- JWTUser should only accepts non-empty-string in the constructor
- createFromPayload should only accepts non-empty-string


VincentLanglet:
Failure are unrelated since I only changed phpdoc

chalasr (approved):
Thanks!

maxhelias (approved):

chalasr:
Thank you @VincentLanglet.

PR #1278, opened 2025-06-04
Remove lcobucci/clock dependency

### LINKED ISSUE #1269: Decouple from lcobucci/clock in favour of psr/clock
Hello everyone,

The problem: my application is currently in the state of having 3 implementations of psr/clock. I want to reduce this quantity to 1.

I noticed, that this bundle uses mostly `$clock->now()` fuctionality from psr/clock and there is only one place, where SystemClock is used (we can get rid of it too). We can easily decouple from `lcobucci/clock` so will be able to use another implementation of `psr/clock`. For example, if you have this bundle v3, you have  `symfony/clock` already installed for sure, because `symfony/security-bundle` v6.3+ depends of it. 

As of system clock, having Psr/Clock interface instead of nullable class in the argument of the service, we will be able to replace it easily if there is such a demand for testing or else.

Is it worth to create a PR and implement all of these there?

VincentLanglet:
Closes https://github.com/lexik/LexikJWTAuthenticationBundle/issues/1269

- `symfony/clock` is already a dependency of `symfony/security-bundle`
- [LooseValidAt](https://github.com/lcobucci/jwt/blob/fc20bfb475929275e279d86eccd0e412396cf108/src/Validation/Constraint/LooseValidAt.php#L17) accepts any clock implementation
- Lcobucci/Clock was implementing `Psr\Clock\ClockInterface` so the change is BC

This allow to remove the dependency


chalasr (approved):

Services/JWSProvider/LcobucciJWSProvider.php
```diff
@@ -44,10 +43,10 @@ class LcobucciJWSProvider implements JWSProviderInterface
     /**
      * @throws \InvalidArgumentException If the given crypto engine is not supported
      */
-    public function __construct(KeyLoaderInterface $keyLoader, string $signatureAlgorithm, ?int $ttl, ?int $clockSkew, bool $allowNoExpiration = false, ?Clock $clock = null)
+    public function __construct(KeyLoaderInterface $keyLoader, string $signatureAlgorithm, ?int $ttl, ?int $clockSkew, bool $allowNoExpiration = false, ?ClockInterface $clock = null)
```
okorneliuk, line 46: Since you touched the constructor, please simplify it by using constructor property promotion and 
`$this->clock = $clock ?? new NativeClock(new \DateTimeZone('UTC'));`

VincentLanglet: constructor property promotion is used nowhere in this bundle.
Let's just minimize the change and keep the consistency.

I would recommend another PR which use constructor property promotion everywhere, if wanted

maxhelias (approved):

VincentLanglet:
Hi @chalasr, tests failure seems unrelated to me, and you approved the PR.

What is needed/Missing for a merge ? 
(same question for https://github.com/lexik/LexikJWTAuthenticationBundle/pull/1277)

Thanks

chalasr:
Thank you @VincentLanglet.

PR #1282, opened 2025-07-15
Fix issue with token generation via DateTimeImmutable instead of int (dependency with lcobucci/jwt ^5.0)

RuslanKononov:
It was problem with generation of token from DTO that implement UserInterface and it was fixed via this changes.

### DIFF
--- a/Subscriber/AdditionalAccessTokenClaimsAndHeaderSubscriber.php
+++ b/Subscriber/AdditionalAccessTokenClaimsAndHeaderSubscriber.php
@@ -4,19 +4,22 @@ namespace Lexik\Bundle\JWTAuthenticationBundle\Subscriber;
 
 use Lexik\Bundle\JWTAuthenticationBundle\Event\JWTCreatedEvent;
 use Lexik\Bundle\JWTAuthenticationBundle\Events;
-use Symfony\Component\Clock\Clock;
+use Psr\Clock\ClockInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Clock\NativeClock;
 
 final class AdditionalAccessTokenClaimsAndHeaderSubscriber implements EventSubscriberInterface
 {
-    /**
-     * @var int|null
-     */
-    private $ttl;
+    private ?int $ttl;
+    private ?ClockInterface $clock;
 
-    public function __construct(?int $ttl)
+    public function __construct(?int $ttl, ?ClockInterface $clock = null)
     {
         $this->ttl = $ttl;
+
+        if (null === $clock) {
+            $this->clock = new NativeClock(new \DateTimeZone('UTC'));
+        }
     }
 
     public static function getSubscribedEvents(): array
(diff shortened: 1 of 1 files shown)


Subscriber/AdditionalAccessTokenClaimsAndHeaderSubscriber.php
```diff
@@ -29,14 +29,16 @@ public static function getSubscribedEvents(): array
 
     public function addClaims(JWTCreatedEvent $event): void
     {
+        $now = new \DateTimeImmutable();
```
maxhelias: You should use `symfony/clock` instead like here : https://github.com/lexik/LexikJWTAuthenticationBundle/blob/3.x/Services/JWSProvider/LcobucciJWSProvider.php#L46-L50

Subscriber/AdditionalAccessTokenClaimsAndHeaderSubscriber.php
```diff
@@ -29,14 +30,17 @@ public static function getSubscribedEvents(): array
 
     public function addClaims(JWTCreatedEvent $event): void
     {
+        $clock = new NativeClock(new \DateTimeZone('UTC'));
```
maxhelias: It's more interesting to inject it into the construct like the example I shared and then make a call like these:
```php
$now = $this->clock->now();
```

Subscriber/AdditionalAccessTokenClaimsAndHeaderSubscriber.php
```diff
@@ -31,12 +35,12 @@ public function addClaims(JWTCreatedEvent $event): void
     {
         $claims = [
             'jti' => uniqid('', true),
-            'iat' => time(),
-            'nbf' => time(),
+            'iat' => $this->clock->now(),
+            'nbf' => $this->clock->now(),
```
maxhelias: Shouldn't these 2 entries have the same instance? Or is it irrelevant?

RuslanKononov: it's irrelevant (it was `time()` before) but you are right, it's more correct to make same nanoseconds time for "initialization" and "not before". maybe it help someone with unit testing

maxhelias (approved):
For me it's ok as the `LcobucciJWSProvider` class hasn't configured the clock injection either.

Subscriber/AdditionalAccessTokenClaimsAndHeaderSubscriber.php
```diff
@@ -4,18 +4,22 @@
...
 {
-    /**
-     * @var int|null
-     */
-    private $ttl;
+    private ?int $ttl;
+    private ?ClockInterface $clock;
 
-    public function __construct(?int $ttl)
+    public function __construct(?int $ttl, ?ClockInterface $clock = null)
```
chalasr, line 16: For the record changing the public API in a patch is forbidden as per semver, as such this will land in a new minor version. For eventual future contributions, consider splitting in 2 PRs: one that fixes the bug at hand. (using the clock) and one that changes the public API (allowing to inject a clock)

PR #1290, opened 2025-10-04
Add support for Symfony 8

chalasr:


mateusfmello:
For the version with Symfony 8 support to be released, is it only necessary for the dependencies to be updated as well?

PR #1295, opened 2025-11-20
Allow to pass extra data when handling success authentication

VincentLanglet:
Hi @chalasr, 

I have a use case when I'd like to return an extra data in the response of the authentication, basically a user "state" (in context of SSO to tell if the user was created or already exists).

In my service I already do my work and then call
```
$this->authenticationSuccessHandler->handleAuthenticationSuccess($user)
```
manually.

Unfortunately I cannot pass aditionnal data this way. Listening the `Events::AUTHENTICATION_SUCCESS` event to modify the data is not a solution, since the listener won't have the information i want to add to the event data.

The best solution I see is, in the same way it's currently possible to pass a specific jwt, to allow passing extra data on the handleAuthenticationSuccess method.

I try to implement it ; WDYT ?

### DIFF
--- a/Security/Http/Authentication/AuthenticationSuccessHandler.php
+++ b/Security/Http/Authentication/AuthenticationSuccessHandler.php
@@ -48,7 +48,7 @@ class AuthenticationSuccessHandler implements AuthenticationSuccessHandlerInterf
         return $this->handleAuthenticationSuccess($token->getUser());
     }
 
-    public function handleAuthenticationSuccess(UserInterface $user, $jwt = null): Response
+    public function handleAuthenticationSuccess(UserInterface $user, $jwt = null, array $data = []): Response
     {
         if (null === $jwt) {
             $jwt = $this->jwtManager->create($user);
@@ -59,8 +59,8 @@ class AuthenticationSuccessHandler implements AuthenticationSuccessHandlerInterf
             $jwtCookies[] = $cookieProvider->createCookie($jwt);
         }
 
-        $response = new JWTAuthenticationSuccessResponse($jwt, [], $jwtCookies);
-        $event = new AuthenticationSuccessEvent(['token' => $jwt], $user, $response);
+        $response = new JWTAuthenticationSuccessResponse($jwt, $data, $jwtCookies);
+        $event = new AuthenticationSuccessEvent(['token' => $jwt] + $data, $user, $response);
 
         $this->dispatcher->dispatch($event, Events::AUTHENTICATION_SUCCESS);
         $responseData = $event->getData();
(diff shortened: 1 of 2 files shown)


VincentLanglet:
Failure seems unrelated.

VincentLanglet:
Pr is rebased @chalasr

chalasr:
Hi @VincentLanglet, I have to admit I find too bad to deviate from the interface signature, but I'm fine doing it as long as there is no other alternative. Can you confirm there is no?

VincentLanglet:
> Hi @VincentLanglet, I have to admit I find too bad to deviate from the interface signature, but I'm fine doing it as long as there is no other alternative. Can you confirm there is no?

Hi, I didn't find any other way currently.

Also, notice the interface is about the 
```
public function onAuthenticationSuccess(Request $request, TokenInterface $token): Response
```
method, while I modify the
```
public function handleAuthenticationSuccess(UserInterface $user, $jwt = null): Response
```
method which is already free from any interface.

The `handleAuthenticationSuccess` was initially created to allow passing a specificy `$jwt` in https://github.com/lexik/LexikJWTAuthenticationBundle/pull/283, therefor adding the `$data` param follow the same philosophy.

chalasr:
Of course, thanks for correcting me. Gonna merge asap and ship it along with Symfony 8 support this week.

chalasr (approved):

VincentLanglet:
> Of course, thanks for correcting me. Gonna merge asap and ship it along with Symfony 8 support this week.

Thanks.

Feel free to ping me if you need help for Symfony 8 support ; I'm interested in it.

chalasr:
Already done in #1290! Help welcome with PHP 8.5 support though #1303

flohw:
@VincentLanglet I add some data to my token via the `Events::JWT_CREATED` event.

```php

    public function onJWTCreated(JWTCreatedEvent $event): void
    {
        $user = $event->getUser();
        if (!$user instanceof User) {
            return;
        }

        $data = $event->getData();
        $data['profile'] = ['foo' => 'bar'];

        $event->setData($data);
    }
```

But maybe that's not what you want?

VincentLanglet:
> @VincentLanglet I add some data to my token via the `Events::JWT_CREATED` event.
> 
> ```
>     public function onJWTCreated(JWTCreatedEvent $event): void
>     {
>         $user = $event->getUser();
>         if (!$user instanceof User) {
>             return;
>         }
> 
>         $data = $event->getData();
>         $data['profile'] = ['foo' => 'bar'];
> 
>         $event->setData($data);
>     }
> ```
> 
> But maybe that's not what you want?

Yes, but this require to know the data to add in your Listener (I already do this some time).
Here I precised 

> Listening the Events::AUTHENTICATION_SUCCESS event to modify the data is not a solution, since the listener won't have the information i want to add to the event data.

And the same issue apply for `Events::JWT_CREATED`.

For my use case I need to add data that only the service calling 
```
$this->authenticationSuccessHandler->handleAuthenticationSuccess($user)
```
knows.

chalasr:
Thank you @VincentLanglet.

PR #1308, opened 2026-04-01
fix: JWTUserProvider stale user cache in long-running runtimes

### LINKED ISSUE #1307: `JWTUserProvider` in-memory cache causes stale roles in long-running PHP runtimes (FrankenPHP)
## Describe the bug

`JWTUserProvider` caches `JWTUser` instances by identifier in a plain PHP `array $cache`:

```php
// Security/User/JWTUserProvider.php
public function loadUserByIdentifierAndPayload(string $identifier, array $payload): UserInterface
{
    if (isset($this->cache[$identifier])) {
        return $this->cache[$identifier]; // ? stale user returned, payload ignored
    }
    $class = $this->class;
    return $this->cache[$identifier] = $class::createFromPayload($identifier, $payload);
}
```

In classic PHP-FPM each request gets a fresh process, so the cache is invisible. In **long-running runtimes** (FrankenPHP worker mode, Swoole, RoadRunner, etc.) the DI container, and therefore the `JWTUserProvider` instance, **persists across requests**. The cache accumulates entries and is never cleared.

**Consequence:** the first request that authenticates with a given `username` value populates the cache. Every subsequent request bearing the same `username` but **different roles/claims** receives the stale user object from the first request, completely ignoring the new JWT payload. This is a **security issue**: users may be granted or denied roles they should not have.

## To Reproduce

1. Deploy a Symfony app with `lexik_jwt` user provider under FrankenPHP (or any long-running runtime).
2. Issue JWT **A** ? `username: "guest"`, `roles: ["ROLE_USER", "ROLE_ADMIN"]`.
3. Make one authenticated request with JWT A ? user is cached as `guest` with `ROLE_ADMIN`.
4. Issue JWT **B** ? `username: "guest"`, `roles: ["ROLE_USER"]` (no admin).
5. Make a request with JWT B ? `JWTUserProvider` returns the cached user from step 3, with `ROLE_ADMIN` still present. The user from JWT B is incorrectly granted admin access.

## Expected behavior

When a JWT carries a different payload than the one used to populate the cache, the provider should create a fresh user rather than returning the stale cached entry. At minimum, the cache should be invalidated between requests in long-running runtimes.

## Proposed fix

Implement `Symfony\Contracts\Service\ResetInterface` on `JWTUserProvider`. Symfony's runtime component automatically calls `reset()` on all services that implement this interface between requests (this is already used throughout Symfony core for exactly this pattern):

```php
use Symfony\Contracts\Service\ResetInterface;

final class JWTUserProvider implements PayloadAwareUserProviderInterface, ResetInterface
{
    // ...

    public function reset(): void
    {
        $this->cache = [];
    }
}
```

This is the minimal, non-breaking change. The cache still provides its within-request deduplication benefit; it is simply cleared at the request boundary, which is the correct behaviour for a stateless JWT authenticator.

## Environment

- `lexik/jwt-authentication-bundle`: current `main`
- Runtime: FrankenPHP worker mode (reproducible with any long-running PHP runtime)
- Symfony: 7.4

ambroisemaupate:
In long-running PHP runtimes (FrankenPHP, RoadRunner, Swoole), the DI container persists across requests, causing JWTUserProvider's in-memory cache to accumulate stale user objects. A user authenticating with a new JWT carrying different roles would receive the cached user from a prior request, leading to incorrect role assignment.

Implement ResetInterface on JWTUserProvider and tag the service with kernel.reset so that Symfony's ServicesResetter clears the cache between requests. The tag is required because the service is registered manually without autoconfigure.

Fixes #1307

### DIFF
--- a/Security/User/JWTUserProvider.php
+++ b/Security/User/JWTUserProvider.php
@@ -3,13 +3,14 @@
 namespace Lexik\Bundle\JWTAuthenticationBundle\Security\User;
 
 use Symfony\Component\Security\Core\User\UserInterface;
+use Symfony\Contracts\Service\ResetInterface;
 
 /**
  * JWT User provider.
  *
  * @author Robin Chalas <robin.chalas@gmail.com>
  */
-final class JWTUserProvider implements PayloadAwareUserProviderInterface
+final class JWTUserProvider implements PayloadAwareUserProviderInterface, ResetInterface
 {
     private string $class;
 
@@ -64,5 +65,10 @@ final class JWTUserProvider implements PayloadAwareUserProviderInterface
     public function refreshUser(UserInterface $user): UserInterface
     {
         return $user; // noop
+    }
+
+    public function reset(): void
+    {
+        $this->cache = [];
     }
 }
--- a/Resources/config/token_extractor.php
+++ b/Resources/config/token_extractor.php
@@ -40,6 +40,7 @@ return static function (ContainerConfigurator $container) {
 
     $services->set('lexik_jwt_authentication.security.jwt_user_provider', JWTUserProvider::class)
         ->private()
+        ->tag('kernel.reset', ['method' => 'reset'])
         ->args([
             '',
         ]);


chalasr (approved):

chalasr:
For eventual future contributions, a test case is much appreciated (follow up PR welcome if you feel brave enough :))

chalasr:
Thank you @ambroisemaupate.

Security/User/JWTUserProvider.php
```diff
@@ -3,13 +3,14 @@
 namespace Lexik\Bundle\JWTAuthenticationBundle\Security\User;
 
 use Symfony\Component\Security\Core\User\UserInterface;
+use Symfony\Contracts\Service\ResetInterface;
```
stof, line 6: For robustness, the composer.json should declare a dependency on `symfony/service-contracts` instead of relying on a transitive dependency (or the interface should be removed, as it is not necessary in this case)

chalasr: Added the requirement in #1309, the interface does no harm since we get this package installed anyway. Thanks

PR #1311, opened 2026-05-29
Add support for AttributesBasedUserProviderInterface

VincentLanglet:
Hi @chalasr 

Symfony provides a AttributesBasedUserProviderInterface
https://github.com/symfony/security-core/blob/a8239abe61dafdd0c01c0b4019138b2855717f97/User/AttributesBasedUserProviderInterface.php#L24
 which seems similar to the existing PayloadAwareUserProviderInterface of the bundle, with the benefit that Symfony has custom check for AttributesBasedUserProviderInterface in his own code base.
 
 So I think that
- This bundle should support AttributesBasedUserProviderInterface the same way than PayloadAwareUserProviderInterface
- Maybe deprecate PayloadAwareUserProviderInterface in favor of AttributesBasedUserProviderInterface

### DIFF
--- a/Security/Authenticator/JWTAuthenticator.php
+++ b/Security/Authenticator/JWTAuthenticator.php
@@ -24,6 +24,7 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
 use Symfony\Component\Security\Core\Exception\AuthenticationException;
 use Symfony\Component\Security\Core\Exception\UserNotFoundException;
+use Symfony\Component\Security\Core\User\AttributesBasedUserProviderInterface;
 use Symfony\Component\Security\Core\User\ChainUserProvider;
 use Symfony\Component\Security\Core\User\UserInterface;
 use Symfony\Component\Security\Core\User\UserProviderInterface;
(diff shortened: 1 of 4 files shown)


chalasr:
Makes sense 👍 deprecation included

VincentLanglet:
> Makes sense 👍 deprecation included

Does something miss for a merge/release then ? :)

chalasr:
In Montréal for SymfonyDay, I'm going to review this early next week :)

Security/Authenticator/JWTAuthenticator.php
```diff
@@ -185,13 +186,19 @@ protected function loadUser(array $payload, string $identity): UserInterface
         if ($this->userProvider instanceof PayloadAwareUserProviderInterface) {
             return $this->userProvider->loadUserByIdentifierAndPayload($identity, $payload);
         }
+        if ($this->userProvider instanceof AttributesBasedUserProviderInterface) {
+            return $this->userProvider->loadUserByIdentifier($identity, $payload);
+        }
```
chalasr: Can we add a runtime deprecation notice using symfony/deprecation-contracts? and test-cover it?

VincentLanglet: I think it's ok now

chalasr:
Thank you @VincentLanglet.

